PR #3, opened 2024-10-01
Minor improvements

SergioGasquez:
- Use Rhai to check conditions
- Deduplicate if statements for `#` and `//`.
- Bump dependencies

## Tests
I've generated a few projects and they seem sane.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -321,12 +321,11 @@ fn process_file(
     scope.push("options", options.to_vec());
 
     // Define a custom function to check if conditions of the options.
-    engine.register_fn(
-        "contains_option",
-        |cond: &str, options: Vec<String>| -> bool {
-            options.iter().any(|option| cond.contains(option))
-        },
-    );
+    let options_clone: Vec<String> = options.iter().map(|v| v.to_owned()).collect();
+    engine.register_fn("option", move |cond: &str| -> bool {
+        let cond = cond.to_string();
+        options_clone.contains(&cond)
+    });
 
     for line in contents.lines() {
         let trimmed: &str = line.trim();
@@ -380,9 +379,7 @@ fn process_file(
             } else {
                 trimmed.strip_prefix("//IF ").unwrap()
             };
-
-            let cond = process_contains_option(cond.to_string(), &mut scope, &engine);
-            let res = engine.eval::<bool>(&cond).unwrap();
+            let res = engine.eval::<bool>(cond).unwrap();
             include.push(res && *include.last().unwrap());
         } else if trimmed.starts_with("#ENDIF") || trimmed.starts_with("//ENDIF") {
             include.pop();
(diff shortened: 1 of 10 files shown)


src/main.rs
```diff
@@ -337,58 +382,55 @@ fn process_file(
         first_line = false;
 
         // that's a bad workaround
-        if line.trim() == "#[rustfmt::skip]" {
+        // Why do we require this, we dont use it (?)
+        if trimmed == "#[rustfmt::skip]" {
```
bjoernQ, line 386: at some point the template contained that because reformatting the code added line-breaks which made the replace-string fail - I guess having it for the future is a good thing

src/main.rs
```diff
@@ -337,58 +382,55 @@ fn process_file(
...
-        } else if line.trim().starts_with("//IF ") {
-            let cond = line.trim().strip_prefix("//IF ").unwrap();
-            if let Some(stripped) = cond.strip_prefix("!") {
-                include.push(!options.contains(&stripped.to_string()) && *include.last().unwrap());
+                trimmed.strip_prefix("//IF ").unwrap()
+            };
+
+            scope.push_dynamic("cond", cond.into());
+
+            let result = engine
```
bjoernQ, line 414: my thinking was we could just pass the whole condition to Rhai and evaluate that

then we could make `contains_option` just operate on one option and we can write conditions like `contains_option("wifi") && (contains_option("alloc) || contains_option("embassy"))`

src/main.rs
```diff
@@ -309,16 +315,28 @@ fn process_file(
...
+    scope.push("options", options.to_vec());
+
+    // Define a custom function to check if conditions of the options.
+    engine.register_fn(
+        "contains_option",
+        |cond: &str, options: Vec<String>| -> bool {
+            options.iter().any(|option| cond.contains(option))
+        },
+    );
+
```
bjoernQ, line 330: I think we can make it easier by registering this function:
```rust
    let options_clone: Vec<String> = options.iter().map(|v| v.to_owned()).collect();
    engine.register_fn(
        "contains_option",
        move |cond: &str| -> bool {
            let cond = cond.to_string();
            options_clone.contains(&cond)
        },
    );
```

and not doing any processing of the condition

src/main.rs
```diff
@@ -337,58 +355,46 @@ fn process_file(
...
+            let cond = if trimmed.starts_with("#IF ") {
+                trimmed.strip_prefix("#IF ").unwrap()
             } else {
-                include.push(options.contains(&cond.to_string()) && *include.last().unwrap());
-            }
-        } else if line.trim().starts_with("//ENDIF") {
+                trimmed.strip_prefix("//IF ").unwrap()
+            };
+
+            let cond = process_contains_option(cond.to_string(), &mut scope, &engine);
```
bjoernQ, line 384: we won't need that with the change of the registered function

src/main.rs
bjoernQ: But then the template would need to have conditions like this `//IF contains_option("alloc")` - probably naming the function just `option` would be more convenient

src/main.rs
bjoernQ: we now can remove this `process_contains_option` and also remove the function

SergioGasquez: Awesome! I definitely overcomplicated things :sweat:

bjoernQ: I probably did a bad job explaining the idea about Rhai ... anyways looks good now!

bjoernQ (approved):
Thanks

PR #4, opened 2024-10-03
Minor improvements - Part 2

SergioGasquez:
- Adds some basic CI
  - We may need to come with a better strategy, but at least we are verifying some stuff already.
  - Example run: https://github.com/SergioGasquez/esp-generate/actions/runs/11158725726
- Adds a `rust-toolchain.toml` file
- Validates, in a very simple way, the arguments
- Unify the `stack-protector` name


bjoernQ (approved):
Thanks

PR #6, opened 2024-10-08
Add an argument for template mode

SergioGasquez:


bjoernQ:
Code looks good so far but I'm wondering if it's really more convenient to have multiple template projects (which will duplicate a lot of Cargo.toml, all of config.toml)

Would both templates always support the same options? If not, we would probably duplicate a lot of that

I don't want to be annoying but maybe we should explore the full path of having two (or later more) templates vs. multiple binaries

SergioGasquez:
I guess it really depends on what we want to support in the near future? If we are going to stay with embassy and blocking for a while, I don't think it's worth splitting them into 2 templates, as it would only reduce a few ifs but duplicate a lot of stuff, and the options if Im not mistaken would be the same as for the blocking template.

But, if we want to add another template, like RTIC, then we should probably split the templates.

bjoernQ:
> I guess it really depends on what we want to support in the near future? If we are going to stay with embassy and blocking for a while, I don't think it's worth splitting them into 2 templates, as it would only reduce a few ifs but duplicate a lot of stuff, and the options if Im not mistaken would be the same as for the blocking template.
> 
> But, if we want to add another template, like RTIC, then we should probably split the templates.

Oh sure - for something totally different like RTIC - yes there another template would make a lot of sense. Haven't thought of RTIC or any other framework t.b.h.

PR #10, opened 2024-10-11
Use current dependencies

bjoernQ:
- use current HAL and other dependencies
- add a `.vscode/settings.json`
- some simplification of the template because we now support expressions as conditions

### DIFF
--- a/template/rust-toolchain.toml
+++ b/template/rust-toolchain.toml
@@ -1,10 +1,9 @@
 [toolchain]
 #IF option("riscv")
-#IF !option("stack-protector")
-#+channel = "stable"
-#ENDIF
 #IF option("stack-protector")
 channel = "nightly"
+#ELSE
+#+channel = "stable"
 #ENDIF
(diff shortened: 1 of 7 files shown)


template/rust-toolchain.toml
```diff
@@ -1,10 +1,10 @@
 [toolchain]
 #IF option("riscv")
 #IF !option("stack-protector")
-channel = "stable"
+#+channel = "stable"
 #ENDIF
 #IF option("stack-protector")
-#+channel = "nightly"
+channel = "nightly"
```
SergioGasquez, line 7: We could do:
```
#IF option("stack-protector")
channel = "nightly"
#ELSE
#+channel = "stable"
#ENDIF
```

SergioGasquez (approved):
LGTM! Only one small nitpick that can be ignored

bjoernQ:
We probably also should run unit tests in CI (now that we have them) but I'm not good at CI stuff

PR #20, opened 2024-10-23
Use `esp-metadata` rather than redefining the `Chip` enum

jessebraham:
Also sorted everything lexicographically, sorry for the noise 😅


SergioGasquez (approved):
LGTM! Thanks

Issue #19, opened 2024-10-23
Code which is generated does not compile


Krensi:
Hi! 

Running following command: `esp-generate --chip esp32c3 -o ble -o wifi -o embassy -o alloc -o probe-rs -o wokwi -o ci xiao-no-std-esp-gen` leads to a project which does not compile. 

Compiler output:
```
WARNING: use --release
  We *strongly* recommend using release profile when building esp-hal.
  The dev profile can potentially be one or more orders of magnitude
  slower than release, and may cause issues with timing-senstive
  peripherals and/or devices.

   Compiling xiao-no-std-esp-gen v0.1.0 (D:\Software\esp32_rust\xiao-no-std-esp-gen)
error[E0277]: the trait bound `EspWifiInitFor: EspWifiTimerSource` is not satisfied
   --> src\bin\async_main.rs:31:9
    |
29  |     let _init = esp_wifi::init(
    |                 -------------- required by a bound introduced by this call
30  |         esp_wifi::EspWifiInitFor::Wifi,
31  |         esp_wifi::EspWifiInitFor::Ble,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `esp_wifi::IntoAnyTimer` is not implemented for `EspWifiInitFor`, which is required by `EspWifiInitFor: EspWifiTimerSource`
    |
    = help: the following other types implement trait `esp_wifi::IntoAnyTimer`:
              Alarm<'_, T, DM, COMP, UNIT>
              AnyTimer
              esp_hal::timer::timg::Timer<T, DM>
    = note: required for `EspWifiInitFor` to implement `EspWifiTimerSource`
note: required by a bound in `esp_wifi::init`
   --> C:\Users\ckren\.cargo\registry\src\index.crates.io-6f17d22bba15001f\esp-wifi-0.10.1\src/lib.rs:406:17
    |
404 | pub fn init(
    |        ---- required by a bound in this function
405 |     init_for: EspWifiInitFor,
406 |     timer: impl EspWifiTimerSource,
    |                 ^^^^^^^^^^^^^^^^^^ required by this bound in `init`

error[E0061]: this function takes 4 arguments but 5 arguments were supplied
   --> src\bin\async_main.rs:29:17
    |
29  |     let _init = esp_wifi::init(
    |                 ^^^^^^^^^^^^^^
...
32  |         timg1.timer0,
[...]

Issue #17, opened 2024-10-16
Double check generated code for Wifi and BLE (especially for async)


bjoernQ:
We currently do what the template did but getting from the generated code to something really working is still unnecessary hard

See https://github.com/esp-rs/esp-hal/discussions/2344#discussioncomment-10956938

Ideally, we should at least setup `Cargo.toml` in a way users could just copy code from the examples (or maybe even include the connection step in the generated Rust code)

bjoernQ:
e.g. for wifi + embassy we should add
```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ esp-wifi = { version = "0.10.1", default-features=false, features = [
     "wifi",
     "esp-alloc",
     "log",
+    "async",
+    "embassy-net",
 ] }
 heapless = { version = "0.8.0", default-features = false }
 smoltcp = { version = "0.11.0", default-features = false, features = [
@@ -39,11 +41,15 @@ smoltcp = { version = "0.11.0", default-features = false, features = [
     "socket-udp",
 ] }
 embassy-executor = { version = "0.6.0",  features = [
+"task-arena-size-12288"
 ] }
 embassy-time     = { version = "0.3.1",  features = ["generic-queue-8"] }
 esp-hal-embassy  = { version = "0.4.0",  features = ["esp32c6"] }
 static_cell      = { version = "2.1.0",  features = ["nightly"] }
 
+embassy-net = { version = "0.4.0", features = [ "tcp", "udp", "dhcpv4", "medium-ethernet"] }
+embedded-io-async = "0.6.1"
+
 [profile.dev]
 # Rust debug is too slow.
 # For debug builds always builds with some optimization
```

To enable the user to just copy + paste the code of the `wifi_embassy_dhcp` example

PR #26, opened 2024-10-29
Check all option combinations in CI

SergioGasquez:
- Adds a `--all-combinations` flag to the `check` command
- Modified the workflow to use that flag when running scheduled, and when running it manually we should be able to select if we want the flag or not
  - I couldnt test this, yet

CI test run: https://github.com/SergioGasquez/esp-generate/actions/runs/11570938398

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -21,6 +21,9 @@ enum Commands {
         /// Target chip to check
         #[arg(value_enum)]
         chip: Chip,
+        /// Verify all possible options combinations
+        #[arg(short, long)]
+        all: bool,
     },
 }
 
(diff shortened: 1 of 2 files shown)


xtask/src/main.rs
```diff
@@ -108,6 +117,29 @@ fn options_for_chip(chip: Chip) -> Vec<Vec<String>> {
...
+                    options.extend(available_options[j].clone());
+                }
+            }
+            result.push(options);
+        }
+        // Filter all the items that contains wifi and ble
+        let result = result
+            .into_iter()
+            .filter(|opts| {
+                !opts.contains(&"wifi".to_string()) || !opts.contains(&"ble".to_string())
```
MabezDev, line 139: What's the purpose of this filter?

SergioGasquez: Currently, when using ble and wifi, we throw an error: https://github.com/esp-rs/esp-generate/blob/main/src/main.rs#L408. I guess its related with https://github.com/esp-rs/esp-generate/issues/17. 

We should decide if we want to close the issue, and keep it like this. Or add support to allow using both options at the same time.

MabezDev: We can fix this later, but we should definitely fix it as they aren't really mutually exclusive.

Probably worth fixing after lands: https://github.com/esp-rs/esp-hal/pull/2301 as it should simplify initializing things

bjoernQ (approved):
Thanks

For the future I have two ideas

- define the options somewhere else so the xtask and main could use the same source of truth since currently we need to duplicate options (and also the xtask doesn't take those optional things into account currently)
- maybe we could have a way to allow generating into an existing directly (but only in CI with some magic ENV-var or something) and just remove the generated files between check runs and speed up checking by not having `cargo check` check a dependency if the features etc. didn change? (Not sure how feasible that is)

xtask/src/main.rs
jessebraham: > We can fix this later, but we should definitely fix it as they aren't really mutually exclusive.

FYI, the TUI does not allow selecting both, this was a temporary fix to resolve #19. See [this comment](https://github.com/esp-rs/esp-generate/issues/19#issuecomment-2431121404).

jessebraham (approved):
Thanks!

PR #29, opened 2024-10-30
Check if the current directory is in a git repo already, and only initialize if not

### LINKED ISSUE #15: Improve git initialization
We should check if we are already in a repo and be more careful with the initialization

jessebraham:
Closes #15

Probably we should set up logging so we can print a warning, but I've opened #28 and we can do this in a separate PR.


bjoernQ (approved):
Thanks!

PR #31, opened 2024-10-31
Miscellaneous cleanup/improvements

jessebraham:
- Add descriptions for command-line arguments
- Use `?` in favour of `unwrap`
- Small cleanup/refactoring

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -256,10 +256,6 @@ fn main() -> Result<(), Box<dyn Error>> {
     fs::create_dir(&project_dir)?;
 
     for &(file_path, contents) in template_files::TEMPLATE_FILES.iter() {
-        if file_path == "Cargo.lock" {
-            continue;
-        }
-
         if let Some(processed) = process_file(contents, &selected, &variables) {
             let file_path = project_dir.join(file_path);
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,8 +1,9 @@
 use std::{
     env,
+    error::Error,
     fs,
     path::{Path, PathBuf},
-    process,
+    process::{self, Command},
 };
 
 use clap::Parser;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -243,57 +252,53 @@ fn main() {
...
-    let dir = path.join(&args.name);
-    std::fs::create_dir(&dir).unwrap();
+    let project_dir = path.join(&args.name);
+    fs::create_dir(&project_dir)?;
 
     for &(file_path, contents) in template_files::TEMPLATE_FILES.iter() {
-        let path = dir.join(file_path);
+        if file_path == "Cargo.lock" {
+            continue;
+        }
```
jessebraham, line 261: Is this check actually required? I don't see `Cargo.lock` listed in `TEMPLATE_FILES`

src/main.rs
```diff
@@ -163,22 +166,28 @@ static CHIP_VARS: &[(Chip, &[(&str, &str)])] = &[
...
 
+    /// Chip to target
     #[arg(short, long)]
     chip: Chip,
 
+    /// Run in headless mode (i.e. do not use the TUI)
     #[arg(long)]
     headless: bool,
 
+    // TODO: Can we list the options and/or point users to some documentation?
```
SergioGasquez, line 180: This is probably worth an issue, readme has the list of options but not sure if that is the best place...

src/main.rs
SergioGasquez: I think this can be removed, the template/Cargo.lock is git ignored.

SergioGasquez (approved):
LGTM! Thanks

src/main.rs
jessebraham: Created #32

PR #34, opened 2024-10-31
Include correct dependencies and features by options

### LINKED ISSUE #17: Double check generated code for Wifi and BLE (especially for async)

bjoernQ:
Fixes #17 

With this it's possible to just copy the code from `wifi_dhcp`, `wifi_ble`, `wifi_embassy_dhcp` and `wifi_embassy_ble` examples (using the correct generator options, minus chip specific `cfg-if`s in the example) and have them working

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -34,6 +34,14 @@ esp-alloc = { version = "0.5.0" }
 #ENDIF
 #IF option("wifi") || option("ble")
 embedded-io = "0.6.1"
+
+#IF option("embassy")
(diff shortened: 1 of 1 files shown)


SergioGasquez (approved):
Thanks!

PR #35, opened 2024-10-31
Assorted cleanup/refactoring of TUI

jessebraham:


src/tui.rs
```diff
@@ -258,37 +232,45 @@ impl Widget for &mut App {
         // Create a space for header, todo list and the footer.
         let vertical = Layout::vertical([
             Constraint::Length(2),
-            Constraint::Min(0),
+            Constraint::Fill(1),
```
jessebraham, line 235: This is functionally the same but better shows the intent, IMO. Pretty needless change otherwise 😅

bjoernQ (approved):
Thanks!

PR #36, opened 2024-10-31
Minor improvements

### LINKED ISSUE #32: Document available options for command-line interface
We allow providing zero or more `--option` arguments when using the CLI, however the available options are not displayed in the help text. We should either find a way to list the options here, or update the description to point users to a list of these options.

bjoernQ:
- list available options in `--help` (closes #32)
- use `target` from `esp-metadata`, avoid duplication

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -141,28 +141,6 @@ static OPTIONS: &[GeneratorOptionItem] = &[
     }),
 ];
 
-static CHIP_VARS: &[(Chip, &[(&str, &str)])] = &[
-    (Chip::Esp32, &[("rust_target", "xtensa-esp32-none-elf")]),
-    (
-        Chip::Esp32c2,
-        &[("rust_target", "riscv32imc-unknown-none-elf")],
-    ),
-    (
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -177,9 +178,14 @@ struct Args {
...
 
-    // TODO: Can we list the options and/or point users to some documentation?
     /// Generation options
-    #[arg(short, long)]
+    #[arg(short, long, help = {
+        let mut all_options = Vec::new();
+        for option in OPTIONS {
+            all_options.extend(option.options());
+        }
+        format!("Generation options: {}",all_options.join(","))
```
SergioGasquez, line 187:
```suggestion
        format!("Generation options: {}",all_options.join(" ,")). For more information regarding the different options. Check the esp-generate Readme.
```

SergioGasquez (approved):
LGTM! Thanks for doing this, just left a comment, as I think it might be useful to point the user at the readme for more info about the different options.

jessebraham (approved):
Thanks!

(Oops, CI broken 😁 )

PR #47, opened 2024-11-17
Editor select

### LINKED ISSUE #42: Suggested optional config files for editors?
It could be nice to let the user select their editor during configuration and get a file similar to the .vscode config, but for other editors such as neovim or helix

This is what helix would look like for example.
`.helix/languages.toml`
```toml
[[language]]
name = "rust"

[language-server.rust-analyzer.config]
check.allTargets = false
cargo.target = "riscv32imac-unknown-none-elf"
```

rmsthebest:
Mostly fixes #42

I will admit I don't know how to configure neovim, but I've added support for the editor I use anyway.
I could have added editors under the `optional` section but i felt as if having deeper nested selection was a worse experience. Can move it if you want me to

I snuck in space to select as well as that is quite a common key in these terminal interfaces.

### DIFF
--- a/template/.helix/languages.toml
+++ b/template/.helix/languages.toml
@@ -1,0 +1,7 @@
+#INCLUDEFILE helix
+[[language]]
+name = "rust"
+
+[language-server.rust-analyzer.config]
+check.allTargets = false
+cargo.target = "riscv32imac-unknown-none-elf"
--- a/template/.helix/languages.toml
+++ b/template/.helix/languages.toml
@@ -4,4 +4,5 @@ name = "rust"
 
 [language-server.rust-analyzer.config]
 check.allTargets = false
+#REPLACE riscv32imac-unknown-none-elf rust_target
 cargo.target = "riscv32imac-unknown-none-elf"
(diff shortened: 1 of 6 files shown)


template/.helix/languages.toml
```diff
@@ -0,0 +1,7 @@
+#INCLUDEFILE helix
+[[language]]
+name = "rust"
+
+[language-server.rust-analyzer.config]
+check.allTargets = false
+cargo.target = "riscv32imac-unknown-none-elf"
```
SergioGasquez, line 7:
```suggestion
check.allTargets = false
#REPLACE riscv32imac-unknown-none-elf rust_target
cargo.target = "riscv32imac-unknown-none-elf"
```

SergioGasquez (commented):
Thanks for the contribution! I think we need to update the target on the helix config file and update the changelog file

template/.helix/languages.toml
rmsthebest: Right, I saw that in the vscode one but for some reason I forgot to think about what it did

jessebraham (approved):
Thanks for the PR! I had done this locally and forgot to open a PR myself 😅 But this appears to match what I did almost exactly so LGTM!

PR #48, opened 2024-11-18
remember position in previous menus

rmsthebest:
This is a very small change, but now that there are two sub-menus it could get annoying if you navigate back and forth if you're moved back to the top all the time.


SergioGasquez (approved):
LGTM! Thanks for making esp-generate better!

PR #49, opened 2024-11-18
Add release process

SergioGasquez:
Automates creating the binary artifacts and publishing the package to crates.io.  The release process is basically copied from `espup`.

We need to build the package before publishing as it needs to create the `template_files.rs`, and for this same reason, we need `--allow-dirty`

- [ ] We need to add the `${{ secrets.CARGO_API_KEY }}` secret

Test release in my fork https://github.com/SergioGasquez/esp-generate/actions/runs/11891708349

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,84 @@
+name: Continuous Deployment
+
+on:
+  release:
+    types: [created]
+
+env:
+  CARGO_TERM_COLOR: always
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  publish-artifacts:
+    name: Generating artifacts for ${{ matrix.job.target }}
+    runs-on: ${{ matrix.job.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - os: macos-latest
+            target: x86_64-apple-darwin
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,88 @@
...
+  publish-artifacts:
+    name: Generating artifacts for ${{ matrix.job.target }}
+    runs-on: ${{ matrix.job.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - os: macos-latest
+            target: x86_64-apple-darwin
+          - os: ubuntu-20.04
```
jessebraham, line 21: Is there any reason to use `20.04` instead of `latest`? Probably doesn't matter but just curious

jessebraham: Ahh okay that rings a bell haha, thanks

jessebraham (approved):

MabezDev (approved):

MabezDev (changes requested):
Can you roll back that last commit, sorry!

MabezDev (approved):

.github/workflows/cd.yml
SergioGasquez: As I did nor recall this, I was just gonna try with ubuntu latest

PR #55, opened 2024-11-21
Ask for confirmation when leaving

SergioGasquez:
- Escape now goes back if on a nested option and quits if on the main menu
- The user needs to confirm quitting the tui


bjoernQ (approved):
Thanks!

SergioGasquez:
The merge queue failed as another CI workflow was triggered and it got canceled, I removed the cancelation thing from the CI as its not acting as in esp-hal

PR #57, opened 2024-11-22
Add missing init

bugadani:


template/src/bin/main.rs
```diff
@@ -44,9 +50,7 @@ fn main() -> ! {
     .unwrap();
     //ENDIF
 
-    let delay = Delay::new();
```
bugadani, line 47: Um, what deleted this 🤔

bjoernQ:
we probably shouldn't do it twice (in the BLE/WIFI case)

bjoernQ (approved):
Thanks

.gitignore
```diff
@@ -1,3 +1,4 @@
 **/target
 src/template_files.rs
 template/Cargo.lock
+/.vscode
```
bjoernQ, line 4: but maybe not in template

bugadani: I specifically checked that it does not ignore the template

bugadani: image

MabezDev:
Could we cut a release now? Not calling `init` is kinda bad :sweat_smile:

PR #58, opened 2024-11-22
UX Improvement

bjoernQ:
This fixes the annoyance when selecting BLE or WIFI when the other was selected before


SergioGasquez (approved):
LGTM! Thanks

Issue #51, opened 2024-11-21
Should we include some very common crates as dependencies by default?


bjoernQ:
While not needed to build the generated code some crates are very common to use - maybe it would be convenient to always have them already included

Currently only `critical-section` comes to my mind but probably there is more?

bjoernQ:
I added `critical-section` in #60 - if anyone identifies more common dependencies we can collect them here - otherwise we can close this at some point in time

MabezDev:
I think we want to add the deps that are needed for the template as it means we'll have to keep updating those too. I suppose stable crates like `critical-section` are fine.

Closing.

PR #60, opened 2024-11-26
Support COEX

### LINKED ISSUE #59: allow option for both ble and wifi coexisting
currently the generator will accept either BLE or Wifi.  Both are possible with the esp_wifi library with the coex feature.  There is probably some technical reason this isn't available yet, but it would be useful to have as an option in future.

bjoernQ:
Closes #59 

In the very beginning it was easier to not support the coex - case (because IF conditions were very inflexible). Now it's easy to support it

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -52,6 +52,9 @@ esp-wifi = { version = "0.11.0", default-features=false, features = [
     #IF option("ble")
     "ble",
     #ENDIF
+    #IF option("wifi") && option("ble")
+    "coex",
+    #ENDIF
     "esp-alloc",
     #IF option("probe-rs")
     #+"defmt",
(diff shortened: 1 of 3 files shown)


template/Cargo.toml
```diff
@@ -92,6 +95,7 @@ embassy-time     = { version = "0.3.1",  features = ["generic-queue-8"] }
 esp-hal-embassy  = { version = "0.5.0",  features = ["esp32c6"] }
 static_cell      = { version = "2.1.0",  features = ["nightly"] }
 #ENDIF
+critical-section = "1.2.0"
```
bjoernQ, line 98: technically not needed but most users will end up needing it (partly resolves #51)

jessebraham (approved):
Thanks!

PR #62, opened 2024-11-28
Add bin section

SergioGasquez:
With this change the generated bin will have the project name instead of `main`, this helps with using esp-template in Wokwi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Update the resulting binary name (#62)
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


bjoernQ:
I don't know much about Wokwi but
```
elf = "target/{{ rust_target }}/debug/{{ project-name }}"
firmware = "target/{{ rust_target }}/debug/{{ project-name }}"
```

shouldn't it use `release` there?


I also thought about a improving the `INCLUDEFILE` functionality to support renaming the target file - might also work if we always name it `main.rs`?

SergioGasquez:
> shouldn't it use release there?

I think its useful to use `debug` as Wokwi is used, many times, to debug. But no hard opinion here.

TBH, I didnt noticed that `wowki.toml` was not working, this PR should also fix it. By making Wokwi things simpler, I meant the `wokwi-builders` we use for Wokwi web version

PR #65, opened 2024-12-03
Verify every dependency is required

### LINKED ISSUE #64: alloc is required for wifi?
Steps to reproduce:

```
$ esp-generate --chip esp32 -o wifi myproject
$ cd myproject
$ cargo check
…
   Compiling esp-hal-procmacros v0.15.0
   Compiling xtensa-lx-rt v0.17.2
error[E0463]: can't find crate for `alloc`
  --> /home/mart-e/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-wifi-0.11.0/src/lib.rs:84:1
   |
84 | extern crate alloc;
   | ^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
The following warnings were emitted during compilation:

warning: esp-wifi@0.11.0: esp-wifi should be built with optimization level 2 or 3 - yours is s.

error: could not compile `esp-wifi` (lib) due to 1 previous error
```

Executing `esp-generate --chip esp32 -o alloc -o wifi myproject` works. I don’t know what `alloc` does but it seems it is needed when enabling `wifi`.

mart-e:
If not, the generated code will not work

Fixes #64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update the resulting binary name (#62)
 
 ### Fixed
+- Verify the required options are provided (#65)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
Thanks!

PR #69, opened 2024-12-26
Use stable toolchain for rust-analyzer on xtensa targets

lasernoises:
It seems that rust-analyzer doesn't support custom toolchains, but the documentation suggests adding an override such that rust-analzer runs on the stable toolchain: https://rust-analyzer.github.io/manual.html#toolchain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Verify the required options are provided (#65)
+- Use stable toolchain for rust-analyzer on xtensa targets (#69)
 
 ### Removed
 
--- a/template/.helix/languages.toml
+++ b/template/.helix/languages.toml
@@ -11,3 +11,7 @@ environment.RUSTUP_TOOLCHAIN = "stable"
 check.allTargets = false
 #REPLACE riscv32imac-unknown-none-elf rust_target
 cargo.target = "riscv32imac-unknown-none-elf"
+#IF option("xtensa")
+check.extraEnv.RUSTUP_TOOLCHAIN = "esp"
+cargo.extraEnv.RUSTUP_TOOLCHAIN = "esp"
+#ENDIF
--- a/template/.helix/languages.toml
+++ b/template/.helix/languages.toml
@@ -2,6 +2,11 @@
 [[language]]
 name = "rust"
 
+#IF option("xtensa")
+[language-server.rust-analyzer]
+environment.RUSTUP_TOOLCHAIN = "stable"
+
+#ENDIF
 [language-server.rust-analyzer.config]
 check.allTargets = false
 #REPLACE riscv32imac-unknown-none-elf rust_target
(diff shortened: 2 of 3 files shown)


bjoernQ:
Apparently the difference this setting makes is using the proc-macro-server from the `stable` toolchain instead of the one from the `esp` toolchain.

The `esp` toolchain is usually released when a new stable toolchain is released (sometimes it takes us a few days) - not sure if this makes a difference then

Do you have an example for when using the `esp` toolchain is worse than using `stable`?

lasernoises:
Sorry, I should have added more detail when opening this.

rust-analyzer doesn't seem to work at all on the `esp` toolchain.

In helix I get the following in the log:

```
2024-12-30T17:53:50.846 helix_lsp::transport [ERROR] rust-analyzer err <- "error: 'rust-analyzer' is not installed for the custom toolchain 'esp'.\n"
2024-12-30T17:53:50.846 helix_lsp::transport [ERROR] rust-analyzer err <- "note: this is a custom toolchain, which cannot use `rustup component add`\n"
2024-12-30T17:53:50.846 helix_lsp::transport [ERROR] rust-analyzer err <- "help: if you built this toolchain from source, and used `rustup toolchain link`, then you may be able to build the component with `x.py`\n"
2024-12-30T17:53:50.846 helix_lsp::transport [ERROR] rust-analyzer err: <- StreamClosed
2024-12-30T17:53:50.846 helix_lsp [ERROR] failed to initialize language server: server closed the stream
```

And I can't add it with rustup. `rustup component add rust-analyzer` gives the following error:

```
error: toolchain 'esp' does not support components
```

I found this Reddit thread which suggested overriding the toolchain: https://www.reddit.com/r/rust/comments/13d2tls/rustanalyzer_and_toolchain_for_esp/
And since 

In this configuration rust-analyzer doesn't seem to run perfectly. It seems to not show all errors, but hover and go-to-definion and so on seem to work for the most part at least.

But maybe I'm missing something. Someone in the Reddit thread suggested that it works in VSCode because it bundles rust-analyzer, but for me that doesn't seem to be the case (I think using the rust-analyzer builds from the VSCode plugin is optional). So maybe the issue doesn't affect everyone on the `esp` toolchain.

Someone also suggested symlinking rust-analyzer from stable into the esp toolchain, but I'd be a bit worried about that causing problems with `esup update` or maybe rustup.

Or is there a different way to get rust-analyzer working on the `esp` toolchain?

bjoernQ:
Thanks for the update!

I'm using VSCode with the RA-plugin installed from the marketplace - there are always some minor problems, but it usually works. Can't say much (or anything) about Helix.

Interestingly `rust-analyzer` isn't contained in the official toolchains for me, too. (`error: Unknown binary 'rust-analyzer.exe' in official toolchain 'stable-x86_64-pc-windows-msvc'.`) And we ship `rust-analyzer-proc-macro-srv` with the esp-toolchain

This doesn't seem to make anything worse so I would be fine with merging this (given it makes RA work for others) - but maybe worth to have this checked by another maintainer first.

SergioGasquez:
I also use RA with VS Code, the changes seem to work for me too, although they report several issues regarding missing xtensa targets: 
image

lasernoises:
Ok, I've made the adjustment. I've also updated the changelog (hope I did that right).

SergioGasquez (approved):
LGTM! Thanks for your investigation and improving the RA config!

PR #70, opened 2025-01-08
Fix dev-container templating

jazzz:
## Problem
The `dev-container` option  (set using `esp-generate --chip esp32 -o dev-container your-project`) does not generate a valid dev-container configuration. The templated parameters are never substituted, which leads to syntax errors, when opening the container. 

This PR updates the dev-container.json file template to use the key/value based substitution found in other files.

### DIFF
--- a/template/.devcontainer/Dockerfile
+++ b/template/.devcontainer/Dockerfile
@@ -54,9 +54,6 @@ RUN if [ -n "${GITHUB_TOKEN}" ]; then export GITHUB_TOKEN=${GITHUB_TOKEN}; fi  \
     --log-level debug \
     --export-file /home/${CONTAINER_USER}/export-esp.sh
 
-# Set default toolchain
-RUN rustup default {{ toolchain }}
-
 # Activate ESP environment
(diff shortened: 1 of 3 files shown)


template/.devcontainer/Dockerfile
```diff
@@ -54,9 +54,6 @@ RUN if [ -n "${GITHUB_TOKEN}" ]; then export GITHUB_TOKEN=${GITHUB_TOKEN}; fi  \
     --log-level debug \
     --export-file /home/${CONTAINER_USER}/export-esp.sh
 
-# Set default toolchain
-RUN rustup default {{ toolchain }}
-
```
jazzz, line 59: Setting the default value for `rustup` appears to irrelevant as the value is overwritten by `rust-toolchain.toml`.

```
rustup show
...
esp (overridden by '/home/esp/project-name/rust-toolchain.toml')
...
```

Removing so there is less confusion for new developers.

SergioGasquez (approved):
LGTM! Thanks for fixing it!

PR #71, opened 2025-01-14
Format dependencies

SergioGasquez:
Removed some lines between dependencies and empty space before `bleps`

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -25,7 +25,6 @@ esp-backtrace = { version = "0.14.2", features = [
     "println",
     #ENDIF
 ]}
-
 esp-hal = { version = "0.22.0", features = [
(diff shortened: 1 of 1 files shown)


SergioGasquez:
> I wonder if there is a formatter we could run on `Cargo.toml` after generation

Maybe we could use https://github.com/tamasfe/taplo, havent tried it before

jessebraham:
> > I wonder if there is a formatter we could run on `Cargo.toml` after generation
> 
> Maybe we could use https://github.com/tamasfe/taplo, havent tried it before

This is used as the formatter by the Even Better TOML extension IIUC, I often add `taplo.toml` to my personal projects. Works fine.

SergioGasquez:
Sounds nice, I might give it a shot!

PR #73, opened 2025-01-15
Update to latest esp-hal releases

SergioGasquez:


template/Cargo.toml
```diff
@@ -70,35 +71,32 @@ esp-wifi = { version = "0.11.0", default-features=false, features = [
...
-    "socket-dhcpv4",
-    "socket-icmp",
     "socket-raw",
-    "socket-tcp",
-    "socket-udp",
 ] }
+edge-dhcp = { version = "0.5.0", git = "https://github.com/bugadani/edge-net", rev = "b72678e15bb7c6e72809df235778bb7b48ac146d" }
+edge-raw = { version = "0.5.0", git = "https://github.com/bugadani/edge-net", rev = "b72678e15bb7c6e72809df235778bb7b48ac146d" }
+edge-nal = { version = "0.5.0", git = "https://github.com/bugadani/edge-net", rev = "b72678e15bb7c6e72809df235778bb7b48ac146d" }
+edge-nal-embassy = { version = "0.5.0", git = "https://github.com/bugadani/edge-net", rev = "b72678e15bb7c6e72809df235778bb7b48ac146d" }
```
jessebraham, line 81: I guess it cannot be avoided for now, but do we have any idea when there will be releases for these packages? Would be nice to depend on published dependencies if possible.

jessebraham (approved):
Thanks!

template/Cargo.toml
bjoernQ: IIRC Ivan said on Matrix it will be "soon"

PR #74, opened 2025-01-16
Use published `edge-*` crates

bjoernQ:
The `edge-*` crates got new releases so we can use them instead of the git revs.

However, I wonder if we should include them at all by default?

e.g. `edge-dhcp` is only useful for AP but I assume that's not very common.

Maybe we should add a hint like `# for more networking protocol support visit https://crates.io/crates/edge-net` instead?

And it seems those crates are async only - so we shouldn't include them (or the hint) if we don't generate for async?

I'm open to suggestions and happy to change this PR accordingly

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -46,7 +46,7 @@ embedded-io = "0.6.1"
 #IF option("embassy")
 embedded-io-async = "0.6.1"
 #IF option("wifi")
-embassy-net = { version = "0.4.0", features = [ "tcp", "udp", "dhcpv4", "medium-ethernet"] }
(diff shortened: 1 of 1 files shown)


SergioGasquez:
> And it seems those crates are async only - so we shouldn't include them (or the hint) if we don't generate for async?

I think we need them even if its not async, otherwise build fails due to the changes on `smoltcp` features iirc.

> Maybe we should add a hint like # for more networking protocol support visit https://crates.io/crates/edge-net instead?

I like the idea of the hint!

SergioGasquez (approved):
LGTM! Thanks

PR #80, opened 2025-01-27
Fix the extraEnv log in the VS Code settings file

### LINKED ISSUE #79: Extra Env option is being added twice
Latest release version.
The chech.extraEnv option gets added twice to the helix and vscode config files.
This is the option being repeated:
```
...
"rust-analyzer.check.extraEnv": {
  "RUSTUP_TOOLCHAIN": "esp"
},
"rust-analyzer.check.extraEnv": {
  "RUSTUP_TOOLCHAIN": "esp"
},
```
Same thing for helix:
```
...
cargo.extraEnv.RUSTUP_TOOLCHAIN = "esp"
cargo.extraEnv.RUSTUP_TOOLCHAIN = "esp"
```

jessebraham:
Fixes #79 (I think)

Not super familiar with this conditional syntax in the templates, but based on local testing at least it seems to work as expected now. @SergioGasquez or @bjoernQ can you confirm please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,19 +13,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed the logic for setting toolchain in `.vscode/settings.json` (#80)
+
 ### Removed
 
 ## [0.2.2] - 2025-01-16
 
 ### Added
+
 - The resulting `Cargo.toml` is now formated with Taplo (#72)
 
 ### Changed
+
 - Update the resulting binary name (#62)
 - Include version of `esp-generate` in the generated code (#67)
 - Use `rustc-link-arg` instead of `rustc-link-arg-bin` (#67)
 
 ### Fixed
+
 - Verify the required options are provided (#65)
 - Use `stable` toolchain for Rust Analyzer on Xtensa targets (#69)
 - Added missing template substitution in `devcontainer.json` (#70)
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
looks good but we probably want to fix `template/.helix/languages.toml` while on it

jessebraham:
> looks good but we probably want to fix `template/.helix/languages.toml` while on it

I'm not really sure what's going on in that file, we seem to be setting the `esp` toolchain 3 separate times 😅

bjoernQ:
> > looks good but we probably want to fix `template/.helix/languages.toml` while on it
> 
> I'm not really sure what's going on in that file, we seem to be setting the toolchain 3 separate times 😅

At least https://github.com/esp-rs/esp-generate/blob/7e4f469466d67b24c083a3b3a31e27d9bc2d404b/template/.helix/languages.toml#L14-L17 looks weird

I guess
`environment.RUSTUP_TOOLCHAIN = "stable"`
and
`check.extraEnv.RUSTUP_TOOLCHAIN = "esp"` 

are intentional, however?

SergioGasquez (commented):
Note that there are no duplications, that we are setting `check.extraEnv` and `cargo.extraEnv` and then `RUSTUP_TOOLCHAIN` environment variable. Dont really know how they affect each other and if we could remove any of them, they were introduced in https://github.com/esp-rs/esp-generate/pull/69 which improved the state of RA. Ill do some investigation and checks

SergioGasquez:
Re reading #69, here is the explanation of why `check.extraEnv` and `cargo.extraEnv`: https://github.com/esp-rs/esp-generate/pull/69#issuecomment-2585210039

bjoernQ:
> Note that there are no duplications, that we are setting check.extraEnv and cargo.extraEnv and then RUSTUP_TOOLCHAIN

oh yeah - it's Monday morning 🙈 ...

jessebraham:
> Note that there are no duplications, that we are setting `check.extraEnv` and `cargo.extraEnv` and then `RUSTUP_TOOLCHAIN` environment variable. Dont really know how they affect each other and if we could remove any of them, they were introduced in #69 which improved the state of RA. Ill do some investigation and checks

Ahh, I should learn how to read I guess 😅

SergioGasquez:
I also though we were duplicating it, until I used ctrl+f :sweat: 

Here is the reason of setting `RUSTUP_TOOLCHAIN`: https://rust-analyzer.github.io/manual.html#toolchain. Regarding `ckeck.extraEnv` and `cargo.extraEnv`... I think setting `cargo.extraEnv` should be enough, as it applies to all cargo commands, `check.extraEnv` only applies to cargo check and extends `cargo.extraEnv`.

I tried an Xtensa project on VsCode and removing `check.extraEnv`, I was still able to make RA work properly with no errors

PR #88, opened 2025-02-04
Check rustc+cargo, espflash, probe-rs

### LINKED ISSUE #63: Error: Could not execute process `espflash flash --monitor ..` No such file or directory
For a fresh new project, created with `esp-generate --chip esp32s3 hello-esp`.

Building with `cargo build` has no errors, but `cargo run` is the one that keeps failing no matter what options I pass. Below is the output of running that

```bash
$ hello-esp git:(main) ✗ cargo run


WARNING: use --release
  We *strongly* recommend using release profile when building esp-hal.
  The dev profile can potentially be one or more orders of magnitude
  slower than release, and may cause issues with timing-senstive
  peripherals and/or devices.

    Finished `dev` profile [optimized + debuginfo] target(s) in 0.10s
     Running `espflash flash --monitor target/xtensa-esp32s3-none-elf/debug/async_main`
error: could not execute process `espflash flash --monitor target/xtensa-esp32s3-none-elf/debug/async_main` (never executed)

Caused by:
  No such file or directory (os error 2)
```

Any thoughts on how to fix this?

bjoernQ:
After generation this will show s.th. like this

image

It's just to hint the user in case of any problems - they can still update / install tools after the project is generated.

Closes #63

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -1,0 +1,88 @@
+use core::str;
+
+use esp_metadata::Chip;
+
+struct Version {
+    major: u8,
+    minor: u8,
+    patch: u8,
+}
+
+enum CheckResult {
+    Ok,
+    WrongVersion,
+    NotFound,
+}
+
+pub fn check(chip: Chip) {
+    let rust_version = get_version(
+        "cargo",
+        if chip.is_xtensa() {
+            &["+esp"]
+        } else {
+            &["+stable"]
+        },
+    );
+
+    let espflash_version = get_version("espflash", &[]);
+    let probers_version = get_version("probe-rs", &[]);
+
+    println!("\nChecking installed versions");
+    print_result("Rust", check_version(rust_version, 1, 84, 0));
+    print_result("espflash", check_version(espflash_version, 3, 3, 0));
+    print_result("probe-rs", check_version(probers_version, 0, 25, 0));
+}
+
+fn print_result(name: &str, check_result: CheckResult) {
+    match check_result {
+        CheckResult::Ok => println!("🆗 {}", name),
+        CheckResult::WrongVersion => println!("🛑 {}", name),
+        CheckResult::NotFound => println!("❌ {}", name),
+    }
+}
+
+fn check_version(version: Option<Version>, major: u8, minor: u8, patch: u8) -> CheckResult {
+    match version {
+        Some(version) => {
+            if version.major >= major && version.minor >= minor && version.patch >= patch {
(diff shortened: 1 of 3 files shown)


bjoernQ:
Not sure this needs a changelog entry?

src/check.rs
```diff
@@ -0,0 +1,88 @@
...
+        } else {
+            &["+stable"]
+        },
+    );
+
+    let espflash_version = get_version("espflash", &[]);
+    let probers_version = get_version("probe-rs", &[]);
+
+    println!("\nChecking installed versions");
+    print_result("Rust", check_version(rust_version, 1, 84, 0));
```
MabezDev, line 31: How do we make sure we keep these up to date?

MabezDev (changes requested):

src/check.rs
SergioGasquez: We could query gh for the latest version. We could copy most of the code from`espup`.

src/check.rs
```diff
@@ -0,0 +1,88 @@
...
+    print_result("espflash", check_version(espflash_version, 3, 3, 0));
+    print_result("probe-rs", check_version(probers_version, 0, 25, 0));
```
SergioGasquez, line 33: Should we check both probe-rs and espflash regardless of the option the user selected?

bjoernQ: I thought about that before but it's just for the user's information - e.g. I usually use espflash for flashing but probe-rs for debugging - so IMHO it's not XOR and ideally at some point a user should have both (e.g. `espflash --erase-chip` is something occasionally needed even when using probe-rs for flashing)

bjoernQ: _but_ maybe showing a red cross here might make people feel uncomfortable - even if they won't need it - not sure

SergioGasquez: We can give it a try and if we see users are confused about this, change it in the futrure

SergioGasquez:
> Not sure this needs a changelog entry?

Imho, its worth an entry, its a new feature that we added

src/check.rs
bjoernQ: I creates #89 to not forget about this

SergioGasquez (approved):

MabezDev (approved):
Thanks!

bjoernQ:
> > Not sure this needs a changelog entry?
> 
> Imho, its worth an entry, its a new feature that we added

Sorry, can't read ... added

PR #90, opened 2025-02-04
Update default probe-rs args

SergioGasquez:


template/.cargo/config.toml
```diff
@@ -2,9 +2,10 @@
 [target.riscv32imac-unknown-none-elf]
 #IF option("probe-rs")
 #REPLACE esp32c6 mcu
-runner = "probe-rs run --chip=esp32c6"
+runner = "probe-rs run --chip=esp32c6 --always-print-stacktrace --no-location --catch-hardfault"
 #ELSE
-#+runner = "espflash flash --monitor"
+#REPLACE esp32c6 mcu
+#+runner = "espflash flash --monitor --chip esp32c6"
```
SergioGasquez, line 8: Added the chip argument as an extra check if users try to flash the wrong target as we do with probe-rs

bjoernQ (approved):
Thanks! This is much better now

PR #94, opened 2025-02-05
Make linker errors more helpful

bjoernQ:
I stumbled upon this "by accident" but thought it might be a good idea - but maybe I'm just a little bit too excited about this so if we don't think it's useful it's fine.

It will look like this

image

While users can't run into it with a freshly generated project they might add and change things which might make them run into it

I guess there are more things that could be helpful - I just didn't want to waste too much time with something we don't think is useful


bjoernQ:
Another situation this can be quite helpful with (in future) is when using esp-wifi with a custom allocator but forgetting to provide the additional required functions

jessebraham (approved):

PR #96, opened 2025-02-10
Support `ELIF` (else-if) in templates

bugadani:
This PR adds support for else-if in templates, which will be useful if we are going to have more options that are mutually exclusive.

The PR also reorders log setup to before HAL initialization, so that the HAL's init process may be able to print logs if such configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - Added a version checker that prints a warn message if not using latest esp-generate version (#87)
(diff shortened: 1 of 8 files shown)


bjoernQ (approved):
Ideally we would add a test case for ELIF - otherwise looks fine

bjoernQ:
> Do we have tests to add it to?

Not a lot of them but especially for the IF/ELSE/ENDIF stuff: https://github.com/esp-rs/esp-generate/blob/de38c80d9932b2717657cd3edd49e86c8df6e8af/src/main.rs#L578-L728

There is probably more we could test ofc

bugadani:
Certainly easier to test this than eyeballing the generated code, thanks. Back to fixing, I go :D

PR #100, opened 2025-02-10
Display help text

bugadani:
This PR implements line counting/wrapping so we don't waste vertical space, as well as adds ability to display extra help lines. Currently used to display what's required, but we can do better with it.

image


SergioGasquez (approved):
LGTM! Nice addition!

PR #101, opened 2025-02-10
Replace automatic option enable/disable with a list of requirements

bugadani:
This may be somewhat of a UX regression, as we no longer enable/disable options automatically, but instead dynamically change whether an item is active or not. We could still reintroduce the automatic requirement selection (which I think may be slightly confusing), but that is a tricky problem to solve now, especially if options get into a requires-requires not cycle.

I'm slightly worried about the regression, because now if we have exclusive options, those don't deselect each other, but the user has to manually toggle them. This also means that the "neither one selected" must be a valid option, although even the previous method didn't handle "option A enables option B enables option C" recursive cases so maybe this is just trading limitations.

The goal of this PR is to enable "Allow unstable HAL features" and "Allow unstable dependencies" options.

After this, we could probably perma-hide features that are not available on the current chip. Having an option suggests it can be enabled somehow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When using `embassy` option, `async_main.rs` file was renamed to `main.rs` (#93)
 - The UI no longer allows selecting options with missing requirements, and does not allow deselecting
   options that are required by other options. (#101)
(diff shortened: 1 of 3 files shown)


bjoernQ (approved):
With the hint showing why a disabled option is disabled I think it's not too bad UX-wise

The other way around (e.g. trying to disable alloc while wifi is enabled) is a bit less ergonomic - but not a big deal with the amount of options we currently have

bugadani:
I'm hoping we can auto-generate the requires/required by helper texts, so that should improve the situation a bit.

PR #103, opened 2025-02-10
Use YAML template description, auto-generate help

bugadani:
image


bugadani:
This PR should also auto-generate the test configurations, just so we don't forget updating them.

PR #106, opened 2025-02-11
Test from yaml

bugadani:
I meant to do this as part of #103 but that was merged before I started :) For current options, the PR returns the same lists for both `all_configurations` and `!all_configurations` as before, but we no longer have to maintain that list.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,33 @@
+pub mod config;
+pub mod template;
+
+pub fn generate_list<S: AsRef<str>>(base: &str, word: &str, els: &[S]) -> String {
+    if !els.is_empty() {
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -0,0 +1,33 @@
+pub mod config;
+pub mod template;
+
+pub fn generate_list<S: AsRef<str>>(base: &str, word: &str, els: &[S]) -> String {
```
bjoernQ, line 4: A less generic name for the function or some docs would be nice

bugadani:
I don't know, I've never been able to run the checks locally.

bjoernQ (approved):
Thanks for taking care of Windows in your other PR

PR #107, opened 2025-02-11
Run generate in debug mode

bugadani:
`esp-generate` fails to compile for some reason on Windows, when ran by the xtask from another directory. I don't really know why, and why this happens with `--release`, but removing that seems to work just fine. `--no-default-features` disables the update checker, which is unnecessary for local/CI testing.


bjoernQ (approved):
Thanks!

PR #109, opened 2025-02-11
Replace deprecated vscode extension in devcontainer config

legas7:
Doing recommended migration of  `crates` to `dependi`  vscode extension.


bjoernQ (approved):
Thanks!

We probably should review the other extensions and also have them for the local VSCode settings

e.g. Wokwi only makes sense when Wokwi support is selected.

Some of the recommendations are pretty much unnecessary IMHO

PR #110, opened 2025-02-11
Maybe CI improvements

bjoernQ:
This might improve CI - we'll see


bjoernQ:
> This might improve CI - we'll see

Hell yeah - it does

PR #111, opened 2025-02-12
Update devcontainer extensions

SergioGasquez:
See https://github.com/esp-rs/esp-generate/pull/109#pullrequestreview-2610890220

### DIFF
--- a/template/.devcontainer/devcontainer.json
+++ b/template/.devcontainer/devcontainer.json
@@ -25,8 +25,20 @@
         "files.watcherExclude": {
           "**/target/**": true
         },
-        "rust-analyzer.checkOnSave.command": "clippy",
-        "rust-analyzer.checkOnSave.allTargets": false,
+        "rust-analyzer.cargo.allTargets": false,
+        //REPLACE riscv32imac-unknown-none-elf rust_target
+        "rust-analyzer.cargo.target": "riscv32imac-unknown-none-elf",
+        //IF option("xtensa")
+        "rust-analyzer.server.extraEnv": {
+          "RUSTUP_TOOLCHAIN": "stable"
+        },
+        "rust-analyzer.check.extraEnv": {
+          "RUSTUP_TOOLCHAIN": "esp"
+        },
+        "rust-analyzer.cargo.extraEnv": {
+          "RUSTUP_TOOLCHAIN": "esp"
+        },
+        //ENDIF
         "[rust]": {
           "editor.defaultFormatter": "rust-lang.rust-analyzer"
         }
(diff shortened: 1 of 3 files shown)


SergioGasquez:
Not sure what to do with https://github.com/esp-rs/esp-generate/blob/main/template/.devcontainer/devcontainer.json#L28-L32, shall we duplicate the settings on `.vscode`?

template/.devcontainer/devcontainer.json
```diff
@@ -35,11 +35,11 @@
         "rust-lang.rust-analyzer",
         "tamasfe.even-better-toml",
         "fill-labs.dependi",
-        "mutantdino.resourcemonitor",
         "yzhang.markdown-all-in-one",
```
bjoernQ, line 38: I don't have this installed - not sure users would need this?

template/.devcontainer/devcontainer.json
```diff
@@ -35,11 +35,11 @@
...
-        "ms-vscode.cpptools",
-        "actboy168.tasks",
-        "Wokwi.wokwi-vscode"
+        //IF option("wokwi")
+        "Wokwi.wokwi-vscode",
+        //ENDIF
+        "ms-vscode.cpptools"
```
bjoernQ, line 42: I have this extension but I often look into ESP-IDF code - probably this makes sense for esp-idf-hal users but I'm not sure this is too useful for esp-hal users?

template/.devcontainer/devcontainer.json
```diff
@@ -35,11 +35,11 @@
         "rust-lang.rust-analyzer",
         "tamasfe.even-better-toml",
         "fill-labs.dependi",
-        "mutantdino.resourcemonitor",
         "yzhang.markdown-all-in-one",
-        "ms-vscode.cpptools",
-        "actboy168.tasks",
-        "Wokwi.wokwi-vscode"
+        //IF option("wokwi")
```
bjoernQ, line 39: ❤️

bjoernQ:
> Not sure what to do with https://github.com/esp-rs/esp-generate/blob/main/template/.devcontainer/devcontainer.json#L28-L32, shall we duplicate the settings on `.vscode`?

mhhh not sure (and I never used a dev-container) but having different settings is not great I guess

btw would you mind to duplicate the recommended extensions in `.vscode/extensions.json`? If we recommend extensions we should also do that for local projects

bjoernQ (approved):
Thanks for taking care of this!

SergioGasquez:
> mhhh not sure (and I never used a dev-container) but having different settings is not great I guess

I updated them to be the same in both files, hopefully those are not changed and stay the same in both places

PR #119, opened 2025-02-13
Implement selection groups

bugadani:
Selection groups behave similar to radio groups in the web: only one option in a group may be selected. They are meant to allow selecting exclusive features (such as log vs defmt), and also allow requiring "one-of-N" options (i.e. defmt requires esp-println, defmt-rtt or rtt-target).

This was pretty hard to get right, evaluating the requirements turned out a bit more complicated than I wanted.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -63,7 +63,7 @@ impl ActiveConfiguration<'_> {
 
     pub fn select(&mut self, option: String) {
         let o = find_option(&option, self.options).unwrap();
-        if !self.requirements_met(o) {
+        if !self.is_option_active(o) {
             return;
         }
         if !Self::deselect_group(&mut self.selected, self.options, &o.selection_group) {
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -38,22 +89,39 @@ impl ActiveConfiguration<'_> {
...
     }
 
-    pub fn requirements_met(&self, option: &GeneratorOption) -> bool {
-        if !option.chips.is_empty() && !option.chips.contains(&self.chip) {
-            return false;
-        }
-
-        // Are this option's requirements met?
-        for requirement in option.requires.iter() {
+    pub fn requirements_met2(&self, requires: &[String]) -> bool {
```
bugadani, line 92: Naming is hard, will clean this up.

bjoernQ (approved):

PR #120, opened 2025-02-13
Allow using defmt or log, allow selecting panic handler

bugadani:


bugadani:
I managed to run `check esp32s3 --all-combinations` locally which gives me some confidence in this PR.

bugadani:
Let's see a full check: https://github.com/esp-rs/esp-generate/actions/runs/13312914162

bjoernQ (approved):
Thanks

Hopefully users won't struggle with the "changed defaults" but that's something we'll learn when they start to use it

bugadani:
Yes indeed. The problem is that if we enable a default selection of options, the TUI becomes inconsistent with the `-o` selection flags. This is because with `-o` we don't (yet) have the ability to disable a selection. I'm not sure if this inconsistency is meaningful or not, I suspect not many use the headless generator, or if they do, they can resolve the issue.

Another slight problem I see is that we would essentially recommend using unstable libraries which feels weird to me.

If these aren't a problem, it's probably not too difficult to enable a recommended set of features by default.

template/template.yaml
```diff
@@ -46,11 +49,69 @@ options:
...
-      it can also be used to interactively debug an application, or run tests on the hardware.
-      Semihosting or RTT-based technologies like defmt-rtt require probe-rs."
+    display_name: Use probe-rs to flash and monitor instead of espflash.
+    help: probe-rs is a debugger that connects to the chips over JTAG. It can be used to flash and
+      monitor, and it can also be used to interactively debug an application, or run tests on the
+      hardware. Semihosting or RTT-based technologies like defmt-rtt require probe-rs.
+    chips:
+      - esp32c6
+      - esp32h2
+      - esp32s3
```
SergioGasquez, line 59: C3 can also come with built-in jtag (?), not sure where it should live

bugadani: I'm not 100% sure, if the default state of the C3 is that the usb jtag is available, then I can move it. But the c3 devkits come with only one USB port that is a serial port. Since this grouping just prints more info, and since I'm unsure, I think the C3 should be good where it is.

bugadani:
> https://rust-lang.github.io/rfcs/2089-implied-bounds.html

Probe-rs can't decode the addresses printed by esp-backtrace, so that option seemed... less useful to me.

PR #129, opened 2025-02-20
Prepare 0.3.0 release

SergioGasquez:


template/src/bin/main.rs
```diff
@@ -54,7 +62,7 @@ fn main() -> ! {
     //ENDIF
 
     //IF option("alloc")
-    esp_alloc::heap_allocator!(72 * 1024);
+    esp_alloc::heap_allocator!(size: 72 * 1024);
```
bjoernQ, line 65: we also need the changes for the async-main

template/src/bin/main.rs
```diff
@@ -67,12 +75,19 @@ fn main() -> ! {
...
     loop {
         //IF option("defmt") || option("log")
         info!("Hello world!");
         //ENDIF
+        //IF option("unestable-hal")
         delay.delay_millis(500);
+        //ELSE
+        let delay_start = Instant::now();
+        while delay_start.elapsed() < Duration::from_millis(500) {}
+        //ENDIF
```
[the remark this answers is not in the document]
SergioGasquez: Tried this before, but clippy doesn't like having an empty look. We could disable this lint, though.

template/src/bin/main.rs
SergioGasquez: Uh! CI is not checking the async-main?? :eyes:

SergioGasquez: We used to check it: https://github.com/esp-rs/esp-generate/actions/runs/13008839425/job/36281730799#step:7:1047

SergioGasquez: The following options are being checked:
- empty, `alloc`, `defmt`, `esp-backtrace`, `log`, `unstable-hal`
We are missing:
- `wifi`, `ble`, `embassy`, `probe-rs`, `flashing-probe-rs/defmt`, `flashing-probe-rs/panic-rtt-target/`, `flashing-espflash/log`, `flashing-espflash/demft`, `flashing-espflash/esp-backtrace`

SergioGasquez: Opened #130, will resolve this

MabezDev: Let's check everything. CI time isn't important here, making sure we have the coverage is.

template/Cargo.toml
```diff
@@ -87,6 +87,7 @@ smoltcp = { version = "0.12.0", default-features = false, features = [
 # for more networking protocol support see https://crates.io/crates/edge-net
 #ENDIF embassy && wifi
 esp-wifi = { version = "0.12.0", default-features=false, features = [
```
MabezDev, line 89: I think we can remove default-features = false here, we manually enable the two default features anyways :D

template/src/bin/main.rs
```diff
@@ -67,12 +71,12 @@ fn main() -> ! {
...
     loop {
         //IF option("defmt") || option("log")
         info!("Hello world!");
         //ENDIF
-        delay.delay_millis(500);
+        let delay_start = Instant::now();
+        while delay_start.elapsed() < Duration::from_millis(500) {}
     }
 
     // for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/v0.23.1/examples/src/bin
```
MabezDev, line 82: We should find a way to programatically derive this from the esp-hal version in the template's Cargo.toml. We can do that another time, for now we should just update this to the beta tag (when it exists). Don't forgot about the async template too.

SergioGasquez: Yes, we should open an issue for it! Shall I update it to `esp-hal-v1.0.0-beta.0` or we keep the old naming here `v1.0.0-beta.0`?

MabezDev: New tag will be `esp-hal-v1.0.0-beta.0`

SergioGasquez:
CI Run with `--all-combination` flags: https://github.com/SergioGasquez/esp-generate/actions/runs/13495625629

PR #131, opened 2025-02-20
Check more

### LINKED ISSUE #130: CI does not cover all the options
The following options are being checked:
- empty, `alloc`, `defmt`, `esp-backtrace`, `log`, `unstable-hal`
We are missing:
- `wifi`, `ble`, `embassy`, `probe-rs`, `flashing-probe-rs/defmt`, `flashing-probe-rs/panic-rtt-target/`, `flashing-espflash/log`, `flashing-espflash/demft`, `flashing-espflash/esp-backtrace`

_Originally posted by @SergioGasquez in https://github.com/esp-rs/esp-generate/pull/129#discussion_r1963645227_

bjoernQ:
Closes #130

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -11,6 +11,7 @@ use esp_generate::{
     template::{GeneratorOptionItem, Template},
 };
 use esp_metadata::Chip;
+use log::info;
 
 #[derive(Debug, Parser)]
 struct Cli {
@@ -32,6 +33,9 @@ enum Commands {
         /// Actually build projects, instead of just checking them
         #[arg(short, long)]
         build: bool,
+        /// Just print what would be tested
+        #[arg(short, long)]
+        dry: bool,
     },
 
     /// Prints all valid combinations of options for a given chip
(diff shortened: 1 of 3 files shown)


xtask/src/main.rs
```diff
@@ -32,6 +36,9 @@ enum Commands {
         /// Actually build projects, instead of just checking them
         #[arg(short, long)]
         build: bool,
+        /// Just print what would be tested
+        #[arg(short, long)]
+        dry: bool,
```
SergioGasquez, line 41: While doing some research for this same issue, I found out that we have an `options` subcommand that does mostly the same, was this on purpose?

SergioGasquez (approved):
LGTM! One small question

xtask/src/main.rs
bjoernQ: true it's definitely the same - personally, I would prefer `dry_run` and remove the `options` command (and rename `dry` to `dry_run`)

What do you think?

SergioGasquez: Sounds good to me!

PR #132, opened 2025-02-24
use manifest path for inclusion of template

MabezDev:


bugadani (approved):

src/main.rs
```diff
@@ -19,10 +19,9 @@ mod check;
 mod template_files;
 mod tui;
 
-static TEMPLATE: LazyLock<Template> = LazyLock::new(|| {
-    let options = include_str!("../template/template.yaml");
-    serde_yaml::from_str(options).unwrap()
-});
+const OPTIONS: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/template/template.toml");
```
bjoernQ, line 22: it still needs `include_str!` - now we are parsing the path as YAML

MabezDev: Still not working sadly :(

src/main.rs
```diff
@@ -19,10 +19,9 @@ mod check;
 mod template_files;
 mod tui;
 
-static TEMPLATE: LazyLock<Template> = LazyLock::new(|| {
-    let options = include_str!("../template/template.yaml");
-    serde_yaml::from_str(options).unwrap()
-});
+const OPTIONS: &str = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/template/template.toml"));
```
bjoernQ, line 22: ah it's yaml not toml

PR #137, opened 2025-03-03
Fix missing defmt feature

bugadani:


bjoernQ (approved):
LGTM

MabezDev:
Should we cut a patch release?

bugadani:
Depends on how much effort that takes, this is not much more than a slight inconvenience I think.

MabezDev:
The release process is fully automated for esp-generate, we just need a PR bumping the ver and then tag the release. Once a release is created, it GHA will build artifacts and upload to crates.io.

PR #139, opened 2025-03-06
Update readme

SergioGasquez:
- Reccomend using `--locked` for installation
- Add note about release artifacts and `cargo-binstall`
- Remove early stage note
- Add note about `--headless` flag


bugadani:
Ah CI isn't running because this just touches the README and that is excluded.

PR #140, opened 2025-03-06
fix: prevent ci jobs from running twice on pull requests

j-g00da:
Currently CI jobs run both on `pull_request` and `push`, which causes them to run twice on each push to a PR. I restricted `on.push.branches` to `main` to prevent that.


MabezDev:
Thank you! I've added the skip-changelog label as this doesn't need to be documented to users. Please drop your last commit and we can merge this.

PR #141, opened 2025-03-06
Add option to enable SSP

### LINKED ISSUE #45: Consider enabling stack-smash protection by default
Since we support it, we could add `"-Z", "stack-protector=all",` to `.cargo/config.toml` to get stack-smash protection (and kind of stack overflow protection in our case). Probably with a hint like we use for forcing stack frames.

Otherwise, the feature is quite hard to discover (but will be mentioned in no-std training in a future PR)

bugadani:
Closes #45

### DIFF
--- a/template/template.yaml
+++ b/template/template.yaml
@@ -49,6 +49,11 @@ options:
       - unstable-hal
 
   - !Option
+    name: stack-smashing-protection
+    display_name: Enable stack smashing protection.
+    help: Requires nightly Rust.
+
+  - !Option
     name: probe-rs
     display_name: Use probe-rs to flash and monitor instead of espflash.
     help: probe-rs is a debugger that connects to the chips over JTAG. It can be used to flash and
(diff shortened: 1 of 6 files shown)


template/template.yaml
```diff
@@ -47,6 +47,11 @@ options:
     requires:
       - unstable-hal
 
+  - !Option
+    name: stack-smashing-protection
+    display_name: Enable stack smashing protection.
+    help: Requires nightly Rust.
```
SergioGasquez, line 53: I think we should add the `chips` field as is only available for riscv devices

bugadani: It is?

bjoernQ: no it's available for Xtensa, too

SergioGasquez: Upps, in my head it was only available for riscv sorry!

SergioGasquez (approved):
LGTM! Thanks

bjoernQ (approved):
Thanks - I think it's a nice feature but we made it very hard to discover in the past

Issue #46, opened 2023-04-27
Add svd files for debugging


SergioGasquez:
See https://github.com/esp-rs/esp-idf-template/issues/96

jessebraham:
We had some brief discussion regarding this yesterday. There are some problems here with regards to ensuring that users have the correct version of the SVD.

We are able to tag the PAC releases and use this tool to download the correct version of the SVD from `esp-pacs`, which at least ensures that the correct file is present when generating a project. However, if/when users update their HAL dependency and a new PAC is used, then the SVD will be out of date, and it's not clear how to resolve this.

We will have to spend some more time thinking about how we want to approach this, or if we want to commit to implementing this feature at all.

jessebraham:
This should probably just be closed, IMO. There is no clear path forward here based on the multiple discussions we've had previously on this topic.

bjoernQ:
I tend to agree - someone who needs them will probably figure out and to most people it's probably not that useful

PR #143, opened 2025-03-07
Generate launch.json

### LINKED ISSUE #18: Generate the launch.json
If we are using `probe-rs`, we should include the `launch.json` with the proper debug config. See https://docs.esp-rs.org/book/tooling/debugging/probe-rs.html#example-launchjson. When using Wokwi simulation, we could also add https://github.com/SergioGasquez/esp-template/blob/main/.vscode/launch.json

bugadani:
Closes #18

Tested with an S3 and a C6, all options enabled. F5 starts a pre-built binary (it's unfortunate that users need to `cargo build` beforehand, but it's a probe-rs-vscode limitation I believe), and there is RTT/Defmt output on both MCUs.

The Xtensa debugging experience is subpar, although that isn't exactly unexpected at this point.


bjoernQ:
Should we add anything to `extensions.json` for this to work?

SergioGasquez:
We could also add support for Wokwi debugging: https://github.com/SergioGasquez/esp-template/blob/main/.vscode/launch.json when using `vscode` and `wokwi`. This requires gdb and its a bit more complex, not sure if its worth adding.

bugadani:
I am honesstly not sure I can add it. I never used wokwi, I don't know how to set it up, and esp-generate won't set up the tools required either.

SergioGasquez:
> I am honesstly not sure I can add it. I never used wokwi, I don't know how to set it up, and esp-generate won't set up the tools required either.

Fair enough, lets leave it out of this pr and we can revisit in the future!

SergioGasquez (approved):
LGTM! Thanks! Hopefully we can solve https://github.com/esp-rs/esp-generate/issues/46 to make the debugging experience better

PR #145, opened 2025-03-07
Provide esp-wifi help

### LINKED ISSUE #135: Add new esp-wifi linker helper
From https://github.com/esp-rs/esp-hal/discussions/3188#discussioncomment-12349675.

I think we can detect these missing symbols and give the advice @bugadani gave in the thread, which is to enable the built-in scheduler (and use an external one like ArielOS).

bugadani:
Closes #135


SergioGasquez (approved):
LGTM! Thanks

PR #146, opened 2025-03-11
Enable the smoltcp feature

bugadani:
Noticed that esp-wifi/smoltcp is not enabled even if smoltcp is generated. Also, embassy-net is only added when embassy is present.


SergioGasquez (approved):
LGTM!

PR #148, opened 2025-03-19
Enable defmt or log on more crates

bugadani:
I'm also removing `heapless` since it's not actually used by the templates.


bjoernQ (approved):
Thanks

jessebraham (approved):

PR #149, opened 2025-03-19
Have a `esp-config` binary

### LINKED ISSUE #12: Add TUI for esp-config
It would be good if the user could configure the project with `esp-config` via TUI

bjoernQ:
Closes #12 

This adds a second binary `esp-config` which can be used to configure options managed by `esp-config`

Testing:
- generate a project via `esp-generate`, patch `esp-config = { git = "https://github.com/esp-rs/esp-hal/" }`  at least
- run the binary e.g. via `cargo run --bin=esp-config -- --path=<your generated and patched project>`

You should get a TUI like this

image

Most probably there are things I haven't thought of but it's a start


jessebraham:
A little concerned that having an `esp-config` binary which is installed from a different package than the existing library with the same name (which is in a different repo) might be a potential source of confusion.

bjoernQ:
> A little concerned that having an `esp-config` binary which is installed from a different package than the existing library with the same name (which is in a different repo) might be a potential source of confusion.

Briefly thought about it but usually users won't interact with the library crate at all - but I'm open to naming suggestions ofc

jessebraham:
Yeah, maybe a non-issue, just felt I should mention it 😅

bjoernQ:
If we can agree on multiple binaries, we could also add more small tools - maybe something like https://github.com/espressif/esp-idf-diag could be candidate

MabezDev (changes requested):
Apologies for not getting to this sooner.

So I started reviewing this (and thinking about naming), but I've realized I don't think this belong here in esp-generate.

This topic spawned discussions of a "meta" tool etc, which could be a nice in the future if we flesh out a proper design for it. For now I think we should do the following:

* Move most of this PR content to the esp-config repo
* Re-use as much of the esp-config types and functions
* In this PR (or a new one) add an option here to install esp-config it as an experimental tool
* I think if it's in the esp-config repo, we can keep the binary name
* We should mention the README for esp-config that only the _library_ is guaranteed by semver, and the binary is experimental.

The TL;DR this is really cool, but lets keep this separate and experimental until we have a better idea of how users will use this, and how it might fold into a meta tool in the future.

PR #154, opened 2025-04-01
Print selected options

bugadani:
Sometimes it's nice to be able to copy-paste most of the command, to either double-check selected options, or to reuse them for testing.


SergioGasquez (approved):
LGTM! Thanks

PR #159, opened 2025-04-07
Open an issue if scheduled fails

SergioGasquez:
There is no easy way to test this, I basically copied: https://github.com/esp-rs/esp-hal/blob/main/.github/workflows/ci_nightly.yml#L46-L60


bugadani (approved):
Minus the label, LGTM

PR #173, opened 2025-05-05
Apply fallback style for quirky terminals

### LINKED ISSUE #172: TUI Odd Selection Highlighting
Using `esp-generate` the highlights for selections are behaving in a strange manner. In the shown screenshot, the red highlight is on a row  different than the selected row in bold "Enable HAL features", which then adds a checkmark on a different row "Enable allocations..." when hitting enter. 

Image

Managing to check/select the proper row follows an odd pattern, and it's not clear what is actually highlighted.

bugadani:
Fixes esp-generate in the vscode terminal. Applies a different color scheme for apple terminal
Closes #172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


bjoernQ (approved):
I can confirm it works for me in vscode now

Issue #169, opened 2025-04-30
Create sane debug profile


MabezDev:
A debug compilation profile which can generating runnable (opt level is high enough esp-wifi etc to run) binaries _with_ decent debug capabilities

MabezDev:
It seems opt-level s gets us in a working state for both WiFi and PSRAM as proven in https://github.com/esp-rs/esp-generate/pull/177.

Closing!

bjoernQ:
But maybe we should adapt our checks in various `build.rs` scripts to not warn about opt-level 's' then?

PR #177, opened 2025-05-23
Update dev profile for debugging with `probe-rs`

JurajSadel:
With
```toml
[profile.dev]
opt-level = "s"
codegen-units = 1
```
I was able to somehow _debug_ examples using `wifi` or `PSRAM`.
image

closes https://github.com/esp-rs/esp-generate/issues/169

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -214,6 +214,7 @@ embedded-test = { version = "0.6.0", features = [
 # Rust debug is too slow.
 # For debug builds always builds with some optimization
 opt-level = "s"
+codegen-units = 1        # LLVM can perform better optimizations using a single thread 
 
 [profile.release]
 codegen-units = 1        # LLVM can perform better optimizations using a single thread


template/Cargo.toml
```diff
@@ -214,6 +214,7 @@ embedded-test = { version = "0.6.0", features = [
 # Rust debug is too slow.
 # For debug builds always builds with some optimization
 opt-level = "s"
+codegen-units = 1        # LLVM can perform better optimizations using a single thread
```
bugadani, line 217: This is terrible for build times and shouldn't actually be needed for debugging. 🤔

SergioGasquez: Maybe we can compare the build times and decide with some numbers?

JurajSadel: Yeah, you are right that it is not required, si I guess I can close this PR. Can I close the https://github.com/esp-rs/esp-generate/issues/169 or do we need to do anything else there?

bugadani: We are requiring `--release` for a few things, and showing that warning is not a great user experience. I think that #169 was about trying to figure out which optimization level works with psram/esp-storage/esp-wifi and at the same time still results in a decent compile time, so that we can remove those warnings.

MabezDev: I don't think we can remove the warning, as we should still check users aren't using opt level 0,1. We can expand checks like https://github.com/esp-rs/esp-hal/blob/dcd55a0ab64f3b2d92ac4f47e23f1f8c1d64f254/esp-wifi/build.rs#L41-L49 to include `"s"` as we now know it works.

JurajSadel:
Closing this as nothing needs to be added here.

Issue #171, opened 2025-05-01
Don't check version of unused tool


bugadani:
Generating a project without probe-rs support still checks the version of probe-rs that is installed. Not sure if this is a problem, but it's slightly weird

bjoernQ:
IIRC the thinking was they might need both (espflash and probe-rs) sooner or later. Maybe depends on how we phrase things in the book 🤔

PR #178, opened 2025-05-23
Explain version check results

### LINKED ISSUE #175: Checking installed version: "❌ Rust"
Hi,

I'm using this tool to generate a config with just defaults. Specifcally 
```
> esp-generate --version
esp-generate 0.3.1
```

After creation the tool prints:
```
> esp-generate --chip=esp32s3 -O . foo 

Checking installed versions
❌ Rust
🆗 espflash
🆗 probe-rs
```

What is check is behind the "❌ Rust" and how can I fix that?

It would be helpful if this information was included in the printout. 

Thanks!

bjoernQ:
Closes #175

I was thinking about giving more details (the required version and/or the actual command executed) but maybe that is too much noise and not too helpful

Anyways just relying on emojis was a bit too brief

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -27,11 +27,16 @@ pub fn check(chip: Chip) {
         },
     );
 
+    let rust_toolchain = if chip.is_xtensa() { "esp" } else { "stable" };
+
     let espflash_version = get_version("espflash", &[]);
     let probers_version = get_version("probe-rs", &[]);
 
     println!("\nChecking installed versions");
-    print_result("Rust", check_version(rust_version, 1, 84, 0));
+    print_result(
+        &format!("Rust ({rust_toolchain})"),
+        check_version(rust_version, 1, 84, 0),
+    );
     print_result("espflash", check_version(espflash_version, 3, 3, 0));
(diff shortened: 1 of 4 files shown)


src/check.rs
```diff
@@ -27,20 +27,27 @@ pub fn check(chip: Chip) {
...
 
 fn print_result(name: &str, check_result: CheckResult) {
     match check_result {
         CheckResult::Ok => println!("🆗 {}", name),
-        CheckResult::WrongVersion => println!("🛑 {}", name),
-        CheckResult::NotFound => println!("❌ {}", name),
+        CheckResult::WrongVersion => {
+            println!("🛑 {} (version requirements are not satisfied)", name)
+        }
+        CheckResult::NotFound => println!("❌ {} (not found)", name),
```
bugadani, line 50: At least let's print the minimum required version, also how to install/update. Also, a missing probe-rs may or may not be a hard error depending on the project requirements, and the message printed should reflect that.

bjoernQ: Also somewhat relates to #171

but yeah - a hint if probe-rs is not really needed might help

bugadani (approved):

PR #181, opened 2025-05-26
Remove label

SergioGasquez:
This was making the CI fail. eg: https://github.com/esp-rs/esp-generate/actions/runs/15233805381/job/42845015708


SergioGasquez:
To be honest, I dont remember if I added the label (I dont think I have), nor resolving your comment, sorry. I should have been more careful before merging it.

Issue #89, opened 2025-02-04
Make sure we reference the right versions of crates, Rust and tools (besides Cargo.toml)


bjoernQ:
We have a couple of places where we reference versions

- MSRV in the (not merged yet) tools version check
- That hint about where to look for examples in the generated main
- `espflash` / `probe-rs` versions in the tools version check

Ideally, we can get the MSRV from the `Cargo.toml`

Not sure about `espflash` and `probe-rs` - while users should use a recent version, they might get away with something older that works 🤷‍♂ but we should make sure we suggest a version we know will work

bugadani:
I'm somewhat confused what this issue is about exactly, and what the definition of done would be if someone were to tackle this.

bjoernQ:
I agree this is probably not an actionable task as is or at least figuring out is part of the task

Could be as simple as having some kind of cfg (in the template.yml ?) where we set those versions instead of having versions set/checked all over the place?

bugadani:
Ah so we should store in the template, what the compatible rust compiler, espflash, probe-rs versions are, as opposed to storing them in the tool source? Makes sense, but for probe-rs the CLI API might change a lot still, there are no guarantees there - but perhaps we can signal a "too new" version somehow.

But where do "the right verions of crates" come into play in this issue? Should we try and look up dependencyversions on crates.io? That just sounds like asking for trouble.

bjoernQ:
> But where do "the right verions of crates" come into play in this issue? Should we try and look up dependencyversions on crates.io? That just sounds like asking for trouble.

That would probably be a problem - we could do the first part and see if we need to do more (but the first part would close this issue - once we see there is a need to do more we can create a new issue)

bugadani:
What else do we have to do here?

PR #189, opened 2025-06-06
Prepare relase 0.4.0

SergioGasquez:
Bumped the MSRV as : `  esp-metadata@0.7.0 requires rustc 1.86.0`

### DIFF
--- a/template/src/bin/async_main.rs
+++ b/template/src/bin/async_main.rs
@@ -124,5 +124,5 @@ async fn main(spawner: Spawner) {
         Timer::after(Duration::from_secs(1)).await;
     }
 
-    // for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/esp-hal-v1.0.0-beta.0/examples/src/bin
+    // for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/esp-hal-v1.0.0-beta.1/examples/src/bin
 }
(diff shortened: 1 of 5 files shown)


template/src/bin/async_main.rs
```diff
@@ -124,5 +124,5 @@ async fn main(spawner: Spawner) {
         Timer::after(Duration::from_secs(1)).await;
     }
 
-    // for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/esp-hal-v1.0.0-beta.0/examples/src/bin
+    // for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/esp-hal-v1.0.0-beta.1/examples/src/bin
```
bugadani, line 127: should we parse esp-hal's version from the template toml and replace the version here with the real one automatically? 🤔

bjoernQ: one of the suggestions in #89

bjoernQ (approved):
Thanks

bjoernQ:
oh we probably should change the checked rust version: https://github.com/esp-rs/esp-generate/blob/d2ab090010b4da8a94148dbcb94e6c1eb47319ce/src/check.rs#L47

bugadani:
Hmm I don't think we broke anything with probe-rs, but I guess it's worth double checking. The later versions have some known issues, so it may be better not to.

PR #191, opened 2025-06-06
Update ubuntu runner

SergioGasquez:
Take 2 due to manually replacing it in #190 instead of search and replace 😓


MabezDev (approved):
Thanks!

PR #211, opened 2025-07-14
Explicitly bump static-cell

SergioGasquez:


bugadani:
I'm not entirely sure what changed the main package lockfile but the template change is OK.

PR #218, opened 2025-07-22
Don't ignore .vscode if vscode options are selected

damienpontifex:
- If the vscode option is selected, my thoughts were that it'd seem likely the user would want the .vscode folder to persist
- As such, conditionally not ignoring that folder if the user has selected vscode option


bjoernQ:
> But IDE settings are often local, personalized, and sometimes can even contain secrets, so it's better to not accidentally just share them.

That's my thinking too and it's why I have it in my global .gitignore - needing a `git add -f .vscode` gives me a second to think about it 😄

PR #219, opened 2025-07-24
Replace rhai with somni

bugadani:
While we don't have a licencing/type system issue here, as we've had with evalexpr in esp-config, we might as well use the same library for both.


bjoernQ (approved):
Thanks!

bjoernQ:
Agree - having a zoo of similar but different dependencies across crates doesn't make sense 😄

Issue #170, opened 2025-04-30
Option to install esp-config tui


MabezDev:
See https://github.com/esp-rs/esp-generate/pull/149#pullrequestreview-2793731568

PR #221, opened 2025-08-05
Check esp-config

bjoernQ:
This adds a check for esp-config -see https://github.com/esp-rs/esp-generate/issues/170

This assumes we have it covered in the book

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -45,80 +45,122 @@ enum CheckResult {
     NotFound,
 }
 
-pub fn check(chip: Chip, probe_rs_required: bool, msrv: Version) {
-    // TODO: check +nightly if needed
-    let rust_version = get_version(
-        "rustc",
-        if chip.is_xtensa() {
-            &["+esp"]
-        } else {
-            &["+stable"]
-        },
-    );
+pub fn check(chip: Chip, probe_rs_required: bool, msrv: Version, requires_nightly: bool) {
+    let rust_toolchain = if chip.is_xtensa() {
+        "esp"
+    } else if requires_nightly {
+        "nightly"
+    } else {
+        "stable"
+    };
 
-    let rust_toolchain = if chip.is_xtensa() { "esp" } else { "stable" };
+    let rust_version = get_version("rustc", &[format!("+{rust_toolchain}").as_str()]);
+
     let rust_toolchain_tool = if chip.is_xtensa() { "espup" } else { "rustup" };
 
     let espflash_version = get_version("espflash", &[]);
 
     let probers_version = get_version("probe-rs", &[]);
+
+    let esp_config_version = get_version("esp-config", &[]);
+
     let probers_suggestion_kind = if probe_rs_required {
         "required"
     } else {
         "suggested"
     };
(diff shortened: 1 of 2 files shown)


src/check.rs
```diff
@@ -45,80 +45,123 @@ enum CheckResult {
...
+    let esp_config_version = get_version("esp-config", &[]);
+
     let probers_suggestion_kind = if probe_rs_required {
         "required"
     } else {
         "suggested"
     };
 
-    println!("\nChecking installed versions");
+    create_check_results(
```
bugadani, line 73: You build a very large string in `create_check_results`, and now you just drop it here without printing it.

src/check.rs
```diff
@@ -243,4 +286,131 @@ espflash 1.7.0"#;
...
+                /*probers_version*/ None,
+                /*esp_config_version*/ None,
+                /*probers_suggestion_kind*/ "required",
+            ),
+            "
+Checking installed versions
+❌ Rust (stable) (not found - use `rustup` to install)
+❌ espflash (not found - see https://crates.io/crates/espflash for installation instructions)
+❌ probe-rs (not found - see https://probe.rs/docs/getting-started/installation/ for how to install (required))
+❌ esp-config (not found - see https://probe.rs/docs/getting-started/installation/ for how to install (installation is optional))
```
bugadani, line 409: Smells like copy-paste

src/check.rs
```diff
@@ -243,4 +289,131 @@ espflash 1.7.0"#;
...
+                /*probers_version*/ None,
+                /*esp_config_version*/ None,
+                /*probers_suggestion_kind*/ "required",
+            ),
+            "
+Checking installed versions
+❌ Rust (stable) (not found - use `rustup` to install)
+❌ espflash (not found - see https://crates.io/crates/espflash for installation instructions)
+❌ probe-rs (not found - see https://probe.rs/docs/getting-started/installation/ for how to install (required))
+❌ esp-config (not found - use `cargo install esp-config --features=tui --locked` to install (installation is optional))
```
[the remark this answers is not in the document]
bjoernQ: Thanks for reminding me this PR exists - I think I have a few more little things to add but sure - this emoji looks a bit scary for an optional tool 👍

src/check.rs
```diff
@@ -45,80 +49,161 @@ enum CheckResult {
...
+            espflash_version,
+            probers_version,
+            esp_config_version,
+            probers_suggestion_kind,
+        )
+    );
+
+    if offensive_cargo_config_check(project_path) {
+        println!(
+            "⚠️ `.config/cargo.toml` files found in upper directories - this can cause undesired behavior. See https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure"
```
MabezDev, line 100:
```suggestion
            "⚠️ `.config/cargo.toml` files found in parent directories - this can cause undesired behavior. See https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure"
```

MabezDev (approved):
LGTM with one nit

PR #226, opened 2025-09-16
Swap linker script order, again

bugadani:
This modified `dhcp` example demonstrates the issue: https://github.com/esp-rs/esp-hal/commit/42b17353b732609873f05768dc4dbbb1dc87a444

If you swap `defmt.x` and `linkall.x`, things stop working. Placing defmt.x last no longer seems to cause a flip-link issue, for some reason.

### DIFF
--- a/template/build.rs
+++ b/template/build.rs
@@ -3,11 +3,12 @@ fn main() {
     //IF option("embedded-test")
     println!("cargo:rustc-link-arg-tests=-Tembedded-test.x");
     //ENDIF
+    println!("cargo:rustc-link-arg=-Tlinkall.x");
     //IF option("defmt")
+    // defmt.x can't be the last linker script, or you'll likely get this error:
+    // `failed to demangle defmt symbol `__global_pointer$`: expected value at line 1 column 1`
     println!("cargo:rustc-link-arg=-Tdefmt.x");
     //ENDIF
-    // make sure linkall.x is the last linker script (otherwise might cause problems with flip-link)
-    println!("cargo:rustc-link-arg=-Tlinkall.x");
 }
 
 fn linker_be_nice() {


wisp3rwind:
I've also just encountered the [linked issue](https://github.com/esp-rs/esp-hal/issues/4149), swapping the linker script order as done here works around the problem.

I've been trying to understand what is really going on, the following is what I have found, and it seems to suggest that a proper fix might be relatively easy (assuming I understand things correctly):

`riscv-rt` provides `__global_pointer$` in the RODATA region: https://github.com/rust-embedded/riscv/blob/38af0edc5ce287802cacd8e67c3e3cd1a48827a0/riscv-rt/link.x.in#L135

However, `esp-hal` doesn't use that, and according to `rg __global_pointer` does not define this symbol anywhere.

In that case, and lacking an `.sdata` section, the linker will default to place the symbol in whatever is first in the ELF; see [the code](https://github.com/llvm/llvm-project/blob/41a2dfc0d77d9ad977d1d36358f979abb3a0928f/lld/ELF/Writer.cpp#L1846) and [ŧhe docs](https://github.com/llvm/llvm-project/blob/41a2dfc0d77d9ad977d1d36358f979abb3a0928f/llvm/docs/RISCVUsage.rst#global-pointer-gp-relaxation-and-the-small-data-limit) as well as [this issue mentioning how the code doesn't fully match the documented behavior](https://github.com/llvm/llvm-project/issues/107069).

This will usually not be a section where it makes any sense. I've checked a bunch of executables with `objdump -h <elf>` to find the first section and look at its symbols with `objdump --section <section> --syms <elf>`, and I've seen `__global_pointer$` end up in `.trap`, `.embedded_test` and `.defmt`. The former two seem to tolerate it, but the latter leads to the parsing error that has been reported.

Thus, I would hazard the guess that this problem would be properly solved by providing the `__global_pointer$` symbol somewhere meaningful (`.data + 0x800`?) and ensure that `gp` is initialized accordingly (but maybe `esp-hal` already invokes the [relevant `riscv-rt` code](https://github.com/rust-embedded/riscv/blob/38af0edc5ce287802cacd8e67c3e3cd1a48827a0/riscv-rt/src/asm.rs#L103)? Although if it does, that would mean that `gp` currently takes a garbage value.). This should also have the advantage of actually allowing `gp` usage, leading to better codegen (fewer `auipc`, assuming the compiler is actually capable of doing so).

wisp3rwind:
> Although if it does, that would mean that `gp` currently takes a garbage value.

image

`gp = 0x800`

PR #228, opened 2025-09-19
fix panic due to padding overflow

vaknin:
### Summary

  - Prevent panic in tui.rs:117 triggered when the terminal shrinks.

### Reproduction

  - Launch esp-generate, enter fullscreen, and enter any MCU and project name, don't confirm yet.
  - Exit fullscreen, open another terminal so the terminal is roughly half of its original size.
  - Press Enter to confirm the project name; observe attempt to subtract with overflow from
  src/tui.rs:117:31.

### Fix

  - Use core::saturating_sub when computing padding so it never goes negative, preserving
  formatting without risking overflow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fixed cases where padding overflow caused panic if terminal size changed (#228)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


MabezDev:
Do you have a reproducer, just to check? When I resize the terminal with the tui running I don't see crashes (Linux, Wayland system).

vaknin:
> Do you have a reproducer, just to check? When I resize the terminal with the tui running I don't see crashes (Linux, Wayland system).

Interesting, I've made a few tests according to the 'Reproduction' section in my PR.
It crashes on Alacritty and Kitty, while not on Xterm.
So my guess would be (probably) Linux only, on some of the terminals.

bugadani:
Also crashes on windows with various errors, I'll retest after merge to see if this fixes all of that.

PR #229, opened 2025-09-23
Add Nix support

Georges760:
~introduce INCLUDE_FROM which fetch from an url using curl~


MabezDev:
> I though the dev-container was another kind of build environment.

Good point, I actually forgot these existed :sweat_smile:. I've filed #235 to remove them.

> It actually make very little sense asking for the user to manually use files from another repo in case they want to have Nix build env support in a README that explain how to do everything else automatically. Better close this PR I guess.

Understood. Thank you again for the PR, I will close it now.

Georges760:
I will continue maintaining my fork as I use it.

PR #234, opened 2025-10-02
Use all dram2_seg for heap

### LINKED ISSUE #232: esp-alloc could use .dram2_uninit section by default when enabled
currently esp-generate adds `.dram2_uninit` section to esp-alloc only if you enable both WiFI and BLE at the same time. As they is no downside to use that section of memory always it is kind wasteful to not make it the default and add extra memory from stack instead when need for COEX of WiFI and BLE

bugadani:
Closes #232


MabezDev (approved):
LGTM, thanks!

I wonder if we should/need to think about how users reference `".dram2_uninit"`, like what if we change it? In esp-hal we have the macros to specify where they need to go, and the macro expands and inserts the correct link section for that version of esp-hal.

PR #236, opened 2025-10-03
Remove devcontainer support

SergioGasquez:


teschmitt:
> > oh, why was this removed?
> 
> See the discussion in [#229 (comment)](https://github.com/esp-rs/esp-generate/pull/229#issuecomment-3361357260)

Although it is discussed there, it is not explained there. Could you please elaborate the decision to remove the dev containers option from esp-generate? No front or anything, I'd just like understand the reasoning behind the removal. Thanks!

PR #237, opened 2025-10-06
Update `esp-generate` to upcoming `rc1` changes

playfulFence:


template/.cargo/config.toml
```diff
@@ -18,6 +18,11 @@ DEFMT_LOG="info"
 ESP_LOG="info"
 #ENDIF
 
+[net]
```
playfulFence, line 21: I need this for `git = "https://github.com/esp-rs/esp-hal.git`, should be gone after new packages are released

playfulFence:
Not sure about the `Changelog` entry though. Let me know if you think it should be changed 😉

bjoernQ:
For the things we need/should change before a release maybe we can add comments in those two files - pretty sure I will forget about the comments in this PR as soon as it's merged

bugadani:
TBH if we need esp-hal to be updated before merging this PR, we should update esp-hal before merging this PR and not bother with comments :)

bjoernQ:
sure - maybe make it a draft  for now

PR #244, opened 2025-10-16
Experimental: use `cargo-batch` to reduce CI build times

### LINKED ISSUE #242: (Try to) use `cargo-batch` to reduce build time

playfulFence:
closes #242 
We probably want to use `m1` runner for this to actually have any impact (?)


playfulFence:
It's either me dancing the `cargo-batch` dance wrong, or @bugadani being right about the fact that it needs to be executed on `m1` runner to have any positive difference in build time

https://github.com/playfulFence/esp-generate/actions/runs/18558236275

PR #246, opened 2025-10-22
Add Neovim editor configuration

robjwells:
This PR adds editor configuration for Neovim via an `.nvim.lua` file in the generated project.

It amends the templating code to support Lua's `--` comment syntax. I've added that support for all directives, though the Neovim configuration only makes use of `replace` and `if-endif`.

By default, Neovim does not enable [project-local configuration](https://neovim.io/doc/user/options.html#'exrc'). I've placed a comment at the top of the config file noting the needed setting, to hopefully avoid future questions about the config not being applied.


robjwells:
Force push just adds the PR number to the changelog.

PR #249, opened 2025-10-31
added custom toolchain name argument

ImplFerris:
espup allows to choose custom name for the toolchain other than "esp". It will be nice esp-generate also allow to specify custom toolchain name.


bugadani:
I agree we need this, and a CLI argument is a good enough starting point. Ideally, though, we should offer the option via the TUI. Since I think the PR already stands on its own, and asking you to come up with a completely new way to enter a string in esp-generate would not be reasonable, could you open an issue to track this feature, please?

Also, could you make the new flag optional, so that the tests can succeed?

PR #258, opened 2025-11-21
fix examples link

Friz64:


bugadani:
Could you also update async_main, please?

bugadani (approved):
Thank you very much!

PR #265, opened 2025-12-03
Interactive toolchain setup

playfulFence:
Add optional `--toolchain-name` CLI argument to choose the Rust toolchain for generated projects. (based on https://github.com/esp-rs/esp-generate/pull/249, I will rebase when it's done, I don't want to steal their code 😅 )

It prefers the default toolchain if it supports the target, otherwise the list of applicable toolchains for this target will appear to user choose from installed toolchains

It ignores generic `stable-*` / `beta-*` / `nightly-*` toolchains and falls back to `esp`/prompts user to choose one of installed

DISCUSSION: should I better default it to `default` toolchain or to `active` toolchain?
also, please pay extra attention to `rust-toolchain.yml` file, not sure if logic there is 100% correct

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -787,7 +787,7 @@ fn toolchains_for_target(target: &str) -> Result<Vec<String>> {
         }
 
         // Use rustc with an explicit +toolchain spec, which works for custom
-        // toolchains like `esp` and `esp-alt`.
+        // toolchains like `esp`.
         let spec = format!("+{}", toolchain);
 
         let targets_list = Command::new("rustc")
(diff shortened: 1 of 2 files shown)


playfulFence:
I will probably just close it and start from scratch more or less

bjoernQ:
I finally tried it and to be honest I hate the current UX

image

It currently offers toolchains which are below the MSRV - even `1.88-x86_64-pc-windows-msv` won't be able to get me something working.

Ironically for e.g. `esp32c3` it doesn't ask me e.g. for a specific nightly or a specific stable toolchain.

If we insist on having a CLI option (which is not a TUI option) we shouldn't prompt if it's not given but just go with our defaults (as of before this PR)

---

I honestly think we need to get back to the drawing board - doing this right need much more thought than we thought 😄 

(BTW if we would offer a TUI option we could dynamically disable options which don't fit other selected options like we do with other options ... but as said: it's apparently not as easy as we thought it is ... at least not if we want to do it right)

Issue #261, opened 2025-11-27
Hint is invisible while using `Light` appearance on MacOS


playfulFence:
I'm talking about this hint:
<img width="803" height="127" alt="Image" src="https://github.com/user-attachments/assets/6c55b993-1401-4097-8561-51e2f402718d" />

When user has `Light` colorway of their MacOS appearance, they see nothing instead of a hint

<img width="781" height="383" alt="Image" src="https://github.com/user-attachments/assets/1a2aae01-135d-487e-afb4-656e78e0834d" />

In fact it exists, it's just blending in with the terminal color

<img width="782" height="95" alt="Image" src="https://github.com/user-attachments/assets/3e869f7b-4471-4daf-9678-172658c676a2" />

[`Ratatui`](https://crates.io/crates/ratatui) probably doesn't support color themes (from what I know), but maybe we can find a suitable color that will do it for both appearance options

PR #266, opened 2025-12-04
Fix footer hint and title text color (`Light` MacOS appearance)

playfulFence:
closes https://github.com/esp-rs/esp-generate/issues/261
before:
image

after: 
image


SergioGasquez (approved):
Thanks for the fix! Maybe I would add a changelog entry

Issue #250, opened 2025-11-04
Feature to select toolchain in TUI


ImplFerris:
I have added the custom toolchain name in the PR #249 . This issue tracks adding a prompt in the TUI version to let users choose existing toolchain.

### Related discussion
See the related comment for context:
https://github.com/esp-rs/esp-generate/pull/249#issuecomment-3481655020

PR #268, opened 2025-12-09
Add an option for custom toolchain and interactive toolchain selection

playfulFence:
closes https://github.com/esp-rs/esp-generate/issues/250

I've created a new category for the `tui` template with a placeholder option, which then will be replaced by installed toolchains.

Toolchain list in the TUI is auto-populated from installed rustup toolchains that actually support the selected chip target and current MSRV of a template project

The `Toolchain:` label in the TUI header shows either the user’s chosen toolchain or that template defaults will be used, this is how workflow works too; if user selected different toolchain in the TUI or via `--toolchain` parameter - selected one will be used, otherwise template defaults from `rust-toolchain.yaml`

https://github.com/user-attachments/assets/b759313c-045f-4f31-8278-fa9a2383ee95

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -8,6 +8,8 @@ use crate::check;
 
 /// Return all installed rustup toolchains that support the given `target`
 /// and meet the given MSRV.
+/// Return all installed rustup toolchains that support the given `target`
+/// and meet the given MSRV.
 fn filter_toolchains_for(target: &str, msrv: &check::Version) -> Result<Vec<String>> {
     let output = match Command::new("rustup").args(["toolchain", "list"]).output() {
         Ok(res) => res,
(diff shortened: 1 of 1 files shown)


SergioGasquez:
We should filter out the channels that do not meet the msrv 
image

src/main.rs
```diff
@@ -651,6 +691,219 @@ fn should_initialize_git_repo(mut path: &Path) -> bool {
...
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    };
```
SergioGasquez: Maybe it will be handy to have a `get_default_toolchain`  function, it will also make `populate_toolchain_category` a bit simpler as it's quite big atm

src/main.rs
```diff
@@ -651,6 +691,219 @@ fn should_initialize_git_repo(mut path: &Path) -> bool {
...
+    if let Some(cli) = cli_toolchain {
+        if !available.iter().any(|t| t == cli) {
+            if chip.is_xtensa()
+                && (cli.starts_with("stable")
+                    || cli.starts_with("beta")
+                    || cli.starts_with("nightly"))
+            {
+                bail!(
+                    "Toolchain `{cli}` is not supported for Xtensa targets; \
+                     please use an ESP toolchain (e.g. `esp`)"
```
SergioGasquez: Maybe we should link to the book https://docs.espressif.com/projects/rust/book/getting-started/toolchain.html#xtensa-devices

src/main.rs
```diff
@@ -651,6 +691,219 @@ fn should_initialize_git_repo(mut path: &Path) -> bool {
...
+                );
+            }
+
+            bail!("Toolchain `{cli}` does not have target `{target}` installed");
```
SergioGasquez: I would also link to the book here https://docs.espressif.com/projects/rust/book/getting-started/toolchain.html#risc-v-devices or suggest the rustup command to install the target

src/main.rs
```diff
@@ -651,6 +691,219 @@ fn should_initialize_git_repo(mut path: &Path) -> bool {
     true
 }
 
+/// Return all installed rustup toolchains that support the given `target`
+/// and meet the given MSRV.
+fn filter_toolchains_for(target: &str, msrv: &check::Version) -> Result<Vec<String>> {
```
SergioGasquez: This function is quite complex, maybe we could add some tests to verify that it works as expected

bjoernQ (approved):
LGTM

Issue #262, opened 2025-11-27
Option to execute tool install commands


MabezDev:
A suggestion from @tom-borcin. We currently print the command the user needs to run to install a required tool, but we could quite easily ask the user if they want to install it now [y/n] and we can execute the command for them.

PR #270, opened 2025-12-11
Add a prompt to install a missing or outdated tool

playfulFence:
closes https://github.com/esp-rs/esp-generate/issues/262

In case tool is not installed or its version is lower than a minimal required version, `esp-generate` will now suggest to run a command which will install a needed tool. 

If `probe-rs` was selected as for flashing/logging/debugging - user will not be prompted to install `espflash` in case it's missing and vice-versa. 

https://github.com/user-attachments/assets/d548a8de-f892-475b-ac76-396f24143314

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -121,7 +121,7 @@ pub fn check(
             &[],
             headless,
             Some(&["cargo", "install", "probe-rs-tools", "--locked"]),
-            Some((0, 25, 0)),
+            Some((0, 30, 0)),
         )
     } else {
         get_version("probe-rs", &[])
@@ -190,7 +190,9 @@ fn create_check_results(
         false,
         &format!("Rust ({rust_toolchain})"),
         check_version(rust_version, msrv.major, msrv.minor, msrv.patch),
-        format!("minimum required version is 1.88 - use `{rust_toolchain_tool}` to upgrade"),
+        format!(
+            "minimum required version is {msrv} - run `{rust_toolchain_tool} update` to upgrade"
+        ),
         format!("not found - use `{rust_toolchain_tool}` to install"),
         true,
         &mut result,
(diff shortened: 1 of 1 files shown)


src/check.rs
```diff
@@ -70,15 +75,63 @@ pub fn check(
         "stable"
     };
 
-    let rust_version = get_version("rustc", &[format!("+{rust_toolchain}").as_str()]);
-
     let rust_toolchain_tool = if chip.is_xtensa() { "espup" } else { "rustup" };
 
-    let espflash_version = get_version("espflash", &[]);
+    let rust_install_cmd: &[&str] = if rust_toolchain_tool == "espup" {
+        &["espup", "install"]
```
SergioGasquez: We dont check if they have `espup` installed iirc

src/check.rs
```diff
@@ -298,6 +353,96 @@ fn offensive_cargo_config_check(path: &Path) -> bool {
...
+    if name == "probe-rs" && cfg!(target_os = "linux") {
+        println!(
+            "💡 On Linux, probe-rs requires additional setup before installation.\n\
+            See https://probe.rs/docs/getting-started/installation/ for details."
+        );
+    }
+
+    println!("Do you want to run `{command_str}` now? [y/N]");
+
+    enable_raw_mode().unwrap();
```
MabezDev: I don't think we should be unwrapping here, let's handle the error locally if we can or propagate the error so we can handle it gracefully and report a proper error

src/check.rs
```diff
@@ -70,15 +75,65 @@ pub fn check(
...
+        get_version("espflash", &[])
+    };
 
-    let probers_version = get_version("probe-rs", &[]);
+    let probers_version = if probe_rs_required {
+        get_version_or_install(
+            "probe-rs",
+            &[],
+            Some(&["cargo", "install", "probe-rs-tools", "--locked"]),
+            Some((0, 25, 0)),
```
MabezDev: This is quite old, cc: @bugadani  are we safe to bump to the latest release?

playfulFence: I think last time I installed `probe-rs` from latest release and it went without a problem 🤷🏼. 

However, it might mean nothing, it changes so quickly, but if @bugadani says it's fine - then we probably are okay to do so

MabezDev: Let's use the latest then

src/check.rs
```diff
@@ -298,6 +359,124 @@ fn offensive_cargo_config_check(path: &Path) -> bool {
...
+        println!(
+            "💡 On Linux, probe-rs requires additional setup before installation.\n\
+            See https://probe.rs/docs/getting-started/installation/ for details."
+        );
+    }
+
+    println!("Do you want to run `{command_str}` now? [y/N]");
+
+    if let Err(err) = enable_raw_mode() {
+        println!(
```
MabezDev: 👌

PR #274, opened 2025-12-15
Infer a working tag from the esp-hal tilde version requirement

### LINKED ISSUE #273: Fix examples link
Currently, after https://github.com/esp-rs/esp-generate/pull/267 we generate: 
```
// for inspiration have a look at the examples at https://github.com/esp-rs/esp-hal/tree/esp-hal-v~1.0/examples
```

Which is an invalid link.

bjoernQ:
Closes #273

We mentioned the link-checker there and I was about to add it but that is only partially related and also we should run the link checker on the generated project - not the unprocessed template - so I'd suggest to split out a new issue about the link checker

### DIFF
--- a/template/Cargo.toml
+++ b/template/Cargo.toml
@@ -21,7 +21,7 @@ test = false
 #ENDIF
 
 [dependencies]
-esp-hal = { version = "~1", features = [
+esp-hal = { version = "~1.0", features = [
     #REPLACE esp32c6 mcu
     "esp32c6",
     #IF option("unstable-hal")


template/Cargo.toml
```diff
@@ -21,6 +21,10 @@ test = false
 #ENDIF
 
 [dependencies]
+#IF false
+# The version which is used to insert the link to examples with a suitable tag
+esp-hal-version-full = "1.0.0"
```
bugadani: I'd say instead of this mess, let's strip the prefix in code, add enough `.0`s to make it a full semver version, and create a new template variable from that. We will always have a relevant tag for a version we release, even if we simplify out the patch version here.

SergioGasquez (approved):
LGTM! Thanks

PR #280, opened 2025-12-20
Add `--disable-emojis` cli argument

TgZ39:
Add `--disable-emojis` option to cli arguments to force the use of the fallback ui elements (\* and >) instead of emojis (✅ and ▶️).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Prompt user to install a missing or outdated tool (#270)
+- Add `--disable-emojis` option
 
 ### Changed
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -79,6 +79,10 @@ struct Args {
     /// Rust toolchain to use (rustup toolchain name; must support the selected chip target)
     #[arg(long)]
     toolchain: Option<String>,
+
+    /// Use ASCII characters to display selected options instead of emojis
+    #[arg(long, action)]
+    disable_emojis: bool,
 }
 
 /// Check crates.io for a new version of the application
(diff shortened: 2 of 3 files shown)


SergioGasquez (commented):
We should also filter the emojis that we use for tools detection, just tried this PR and I still get emojis:
```
     Running `target/release/esp-generate --disable-emojis`
> Select your target chip: esp32c3
> Enter project name: asdad
Selected options: --chip esp32c3 -o nightly-x86_64-unknown-linux-gnu
[2025-12-23T09:22:16Z WARN ] Current directory is already in a git repository, skipping git initialization
🛑 esp-config is not installed or is below the required version.
Do you want to run `cargo install esp-config --features=tui --locked` now? [y/N]

Checking installed versions
🆗 Rust (stable): 1.92.0
🆗 espflash: 4.3.0
🆗 probe-rs: 0.30.0
💡 esp-config (not found - use `cargo install esp-config --features=tui --locked` to install (installation is optional))

⚠️ `.config/cargo.toml` files found in parent directories - this can cause undesired behavior. See https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
```

TgZ39:
I think removing emojis completely is a good idea. Instead of using emojis we could display warnings, errors and suggestions using colored text which is supported by more terminals and is still understandable if the terminal doesn't support colors.

TgZ39:
@SergioGasquez That sounds like something I could do. Should the no emoji mode just replace emojis with fitting ASCII characters or should it also use colored text to make it more clear?

SergioGasquez:
>  Should the no emoji mode just replace emojis with fitting ASCII characters or should it also use colored text to make it more clear?

IMHO, in the no-emoji mode, we shouldn't use ASCII characters but maybe colored text could help make things more clear

PR #283, opened 2025-12-27
Add module selector

### LINKED ISSUE #277: Proposal: add module selector
To resolve issues like https://github.com/esp-rs/esp-hal/issues/4521, we could add a module/chip selector. The default value would be "Generic/unknown", but users could pick their specific one if they know what they are working with. The ways it could modify code generation:

- We could generate `let _ = peripherals.GPIOn;` lines and a comment about reserved GPIOs
- We could configure octal PSRAM automatically
- possibly more

KushalMeghani1644:
This PR tends to resolve #277

This PR adds a module selector to esp-generate that allows users to specify their exact ESP module. When a module is selected, the generated code automatically reserves GPIOs that are connected to flash/PSRAM, preventing accidental use that could crash the MCU.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -186,17 +186,6 @@ fn main() -> Result<()> {
         bail!("Directory already exists");
     }
 
-    // Now we filter out the incompatible options, so that they are not shown and they also don't
-    // screw with our position-based data model.
-    let mut template = TEMPLATE.clone();
-    remove_incompatible_chip_options(chip, &mut template.options);
-
-    // Populate module category early so it can be validated
-    module_selector::populate_module_category(chip, &mut template.options);
-
-    // Validate options against the populated template
-    process_options(&template, &args)?;
-
     let versions = cargo::CargoToml::load(
         TEMPLATE_FILES
             .iter()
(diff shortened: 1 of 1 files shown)


KushalMeghani1644:
Great! the CI passes... this PR is now ready for review :)

xtask/src/main.rs
```diff
@@ -14,7 +14,7 @@ use esp_metadata::Chip;
 use log::info;
 
 // Unfortunate hard-coded list of non-codegen options
-const IGNORED_CATEGORIES: &[&str] = &["editor", "optional", "toolchain"];
+const IGNORED_CATEGORIES: &[&str] = &["editor", "module", "optional", "toolchain"];
```
bugadani: This will need to be reverted, all code needs to be tested that esp-generate can generate.

KushalMeghani1644: Hey! @bugadani thanks for your explaination... and I agree that its too early to push this new feature until the team has decided if its worth it or not!

Also I understand your point about the `Generic` option, using “no modules selected” as the generic case is cleaner and avoids redundancy. 

I’m happy to pause work on this PR for now and wait until there’s a clearer decision on the feature itself. If it’s decided to move forward later, I’d be glad to help with implementing proper CI coverage for the generated code.

Maybe it makes sense to continue the discussion on the original issue you opened about this feature?

KushalMeghani1644:
Ugh, messed up Ig, thus I am converting it into a draft for now

PR #289, opened 2026-01-14
add module selector

KushalMeghani1644:
This PR tends to resolve #283

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -186,17 +186,6 @@ fn main() -> Result<()> {
         bail!("Directory already exists");
     }
 
-    // Now we filter out the incompatible options, so that they are not shown and they also don't
-    // screw with our position-based data model.
-    let mut template = TEMPLATE.clone();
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -404,6 +410,26 @@ fn main() -> Result<()> {
...
+    if let Some(ref module_name) = selected_module {
+        if let Some(module) = esp_generate::modules::find_module(module_name) {
+            // Only set module-selected if there are GPIOs to reserve,
+            // otherwise the generated code would have unused `peripherals` variable
+            if !module.reserved_gpios.is_empty() {
+                selected.push("module-selected".to_string());
+                if module.octal_psram {
+                    selected.push("octal-psram".to_string());
+                }
+                let reserved_gpio_code = module
```
MabezDev: We also need to generate a comment above these reserved GPIOs to say why they're reserved (for use with PSRAM/FLASH etc).

PR #293, opened 2026-01-15
Add options to list and get a detailed explanation of options

### LINKED ISSUE #287: Improve accessibility
Would it be possible to specify features with commandline flags to avoid the TUI menu for selecting them?

I am blind and the menu doesn't work too well with my screen reader making it difficult to select the features I want or to know what features I have selected or what is selected by default.
If there is already a way to do this, could it be put in the readme as I could not find a way to do so from a quick read over it.

bugadani:
Closes #287 by introducing `esp-generate list-options` and `esp-generate explain <option>`. These commands describe each option and their dependencies. The module and toolchain options are not included in this yet.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -162,7 +162,14 @@ impl SubCommands {
 
                         if !option.requires.is_empty() {
                             help_text.push_str(" Requires: ");
-                            help_text.push_str(&option.requires.join(", "));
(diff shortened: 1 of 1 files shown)


SergioGasquez (approved):
LGTM! Just gave it a try and it looks like a good improvement to me! I would still wait for some more useful feedback before merging it

PR #303, opened 2026-02-09
Allow using a different xtensa toolchain name

SergioGasquez:
- Updates the IDE configs to use the proper Xtensa toolchain name
- Updates the Rust check to use the proper Xtensa toolchain name


MabezDev (approved):
Thanks!

Issue #167, opened 2025-04-28
Seperate template from binary


MabezDev:
We've discussed this a few times, but I think we need to find a consensus as we'll keep discussing it over and over.

### History

`cargo-generate` uses a separated template system, where the templates are independent from the binary.

### The status quo

In esp-generate, we are bundling the template with the binary. This has some benefits, and some downsides which we'll need to weigh up here.

---

### Advantages

* esp-generate, will _always_ valid project as the template is a snapshot of the ecosystem at a given time
* Template issues are typically easier to debug, just install the users version of esp-generate and input their config
* Project generation works offline

### Disadvantages

* Requires updating esp-generate often
* If a user doesn't update (or misses the fact they have to), they don't get the latest template

### Advantages

* One time install (sort of, there can/will be bugs)
* Might allow for custom templates (think 3rd party templates from ecosystem)

### Disadvantages

* With `cargo-generate`, we struggled to easily repro generation issues
  * This is probably solvable by embedding the git hash of the template in the generated project, and having the option to generate from a git hash 
* Requires the internet to generate projects

bugadani:
> Requires the internet to generate projects

cargo already needs internet connection to download dependencies. esp-generate uses internet connection to check for updates. The user probably should have the option to refer to a template on their file system, but otherwise do we have some requirement against internet connections?

MabezDev:
> > Requires the internet to generate projects
> 
> cargo already needs internet connection to download dependencies. esp-generate uses internet connection to check for updates. The user probably should have the option to refer to a template on their file system, but otherwise do we have some requirement against internet connections?

No requirements, I was just listing pros/cons as I saw them. I don't think requiring an internet connection is unreasonable at all.

bjoernQ:
We probably should provide some way to not rely on an internet connection - I guess pointing to a local template is good enough for that

And a way to choose an older version ofc

MabezDev:
Imo, the bundling of the template is working well and separation doesn't offer enough benefits to move away from our current approach.

PR #311, opened 2026-03-31
Add --bsp option to use external Board Support Package templates

Hahihula:
This is a proposal PR open to discussion.

Adds a new --bsp (Board Support Package) CLI option that allows specifying
an external template directory. This enables using pre-configured
templates generated from esp-bsp JSON files, providing board-specific
pin locations and hardware information.

The feature supports:
  - Local directory paths (with tilde expansion for home directory)
  - Runtime template parsing when using external BSP

This is a prerequisite for implementing an esp-bsp style
board support ecosystem where templates can be versioned and shared
separately from the main esp-generate tool in separate git repository.

The idea is a tool to generate BSP templates from esp-bsp JSON files
and esp-generate base template will be created.

Remains to be solved:
  * chip selection should come from the BSP template

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ serde           = { version = "1.0.228", features = ["derive"] }
 serde_yaml      = "0.9.33"
 toml_edit       = "0.23.9"
 indexmap = "2.13.0"
+walkdir         = "2.5.0"
+quote           = "1.0.43"
 
 [build-dependencies]
 quote   = "1.0.43"
--- a/build.rs
+++ b/build.rs
@@ -1,53 +1,9 @@
+mod template_embed {
+    include!("src/template_embed.rs");
+}
+
 fn main() {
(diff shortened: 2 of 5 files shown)


bugadani:
I like the idea, external templates open up some interesting possibilities. This PR is aA bit on the hacky side - if we do anything of this kind, we should start by turning the built-in template into a less special template pack first, one that we can easily swap out with an external one later.

(I have to add, one form of that idea was rejected in https://github.com/esp-rs/esp-generate/issues/167 )

Hahihula:
> I like the idea, external templates open up some interesting possibilities. This PR is aA bit on the hacky side - if we do anything of this kind, we should start by turning the built-in template into a less special template pack first, one that we can easily swap out with an external one later.
> 
> (I have to add, one form of that idea was rejected in #167 )

My idea was that this way, I am not imposing any more maintenance burden on the esp-generate repo, making everything backwards compatible, while opening the door to more experiments with the bsp-templates generation. 
And this was the least hacky solution I came up with. :-)

Issue #306, opened 2026-03-18
Unable to un-select the flashing method after options for the method were configured


playfulFence:
<img width="893" height="262" alt="Image" src="https://github.com/user-attachments/assets/635b34b1-7e26-4740-8a02-667c121cad6b" />

If I select `Use probe-rs instead of espflash`, then enable some options in `Flashing, logging and debugging (probe-rs)`, I can't switch from `probe-rs` to `espflash` anymore

And vice-versa, if going the opposite direction but at least `Use probe-rs instead of espflash` becomes unavailable. 

PS: if you only select `defmt` printing, you can still switch between both, as this option is the same for them

Desired state:
1) I'd like to be always able to switch between two, with cancelling selected options for the method being switched (preferred, I guess)
2) Make it obvious that switching is impossible after options for the flashing method were selected (e.g. warning in the footnote

bjoernQ:
We might want to re-consider the way to switch the flashing method - while working fine, the current situation is kind of a work-around already

bjoernQ:
One way to mitigate things w/o any code changes (just changing the template-yml) might be to have these options in just one category e.g.

<img width="1940" height="273" alt="Image" src="https://github.com/user-attachments/assets/cce87601-c864-45fe-81bd-3ad71c18cf98" />

Still not ideal but at least makes it easier to understand what is needed to un-select "probe-rs"

playfulFence:
Closed in https://github.com/esp-rs/esp-generate/pull/312

PR #312, opened 2026-04-01
Preserve per-method flashing options when switching probe-rs and espflash

KushalMeghani1644:
Implement flashing-method state memory in the TUI so switching between `probe-rs` and `espflash` restores each method's previous method-specific selections.

- Keep shared options (like `defmt`) selected across method switches.

- Update flashing-related help text in `template/template.yaml` to explain the new switching behavior.

This PR tends to resolve issue: #306


bugadani:
Please don't hard-code assumptions about the template options

KushalMeghani1644:
@bugadani thanks for clarifying, I’ll rework it to be template-driven: derive switchable method options and their dependent option sets from `requires` (`<method>` `/ !<method>`) at runtime, so no option names are hard-coded. I’ll also update tests to validate the generic behavior, are you fine with that?

xtask/Cargo.lock
playfulFence, line 1: Could you please exclude/remove `Cargo.lock` from the PR? 😅

template/template.yaml
```diff
@@ -112,6 +114,8 @@ options:
   - !Category
     name: flashing-probe-rs
     display_name: Flashing, logging and debugging (probe-rs)
+    help: Only options specific to probe-rs are shown here. They are remembered when switching
```
playfulFence: I don't think these under-the-hood details should be introduced to the user

template/template.yaml
```diff
@@ -134,6 +138,8 @@ options:
   - !Category
     name: flashing-espflash
     display_name: Flashing, logging and debugging (espflash)
+    help: Only options specific to espflash are shown here. They are remembered when switching
```
playfulFence: I don't think these under-the-hood details should be introduced to the user

template/template.yaml
```diff
@@ -88,7 +88,8 @@ options:
     display_name: Use probe-rs to flash and monitor instead of espflash.
     help: probe-rs is a debugger that connects to the chips over JTAG. It can be used to flash and
       monitor, and it can also be used to interactively debug an application, or run tests on the
-      hardware. Semihosting or RTT-based technologies like defmt-rtt require probe-rs.
+      hardware. Semihosting or RTT-based technologies like defmt-rtt require probe-rs. Switching
```
playfulFence: I don't think these under-the-hood details should be introduced to the user 🤔

PS: I meant `Switching between probe-rs and espflash restores the last flashing selections for each method.` a bit lower than a snippet shows, sorry 😄

template/template.yaml
```diff
@@ -99,7 +100,8 @@ options:
...
+      between probe-rs and espflash restores the last flashing selections for each method.
```
playfulFence: I don't think these under-the-hood details should be introduced to the user

playfulFence (changes requested):
When just opened an application, selected a some espflash options, then went back to main menu, the `probe-rs` selection is grayed out and the help footer says that it's `Disabled by `log` and `esp-backtrace`, but in fact it's clickable and works as expected. After a second try, everything is shown as expected

Image

playfulFence:
Hmmm, something weird is going on here 😅 

When I choose some `espflash` options, switch to probe-rs and switch back - no `espflash` options saved. 

But if I choose `espflash` options, switch to `probe-rs`, choose some `probe-rs` options and switch back - `espflash` options are restored 🤔 

It's easier to show than to describe 😅 

https://github.com/user-attachments/assets/026ac274-79dc-42f4-abfe-4ca470e93a1a

KushalMeghani1644:
Hey @playfulFence I am sorry for the inconviniences, I have fixed the bug you were facing :)

https://github.com/user-attachments/assets/1497852f-da30-4ae7-9fbc-2d8f0e616728

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Preserve per-method flashing options when switching probe-rs and espflash
 - Add module selector
```
playfulFence:
```suggestion
- Preserve per-method flashing options when switching probe-rs and espflash (#312)
- Add module selector (#289)
```

xtask/Cargo.lock
playfulFence: This one is still not excluded from the diff, could you revert `Cargo.lock` entirely please?

src/tui.rs
```diff
@@ -56,7 +59,123 @@ impl Repository {
...
+        method_option: &str,
+        method_option_selected: bool,
+    ) -> bool {
+        if method_option_selected {
+            Self::option_requires(option, method_option)
+        } else {
+            let requirement = format!("!{method_option}");
+            Self::option_requires(option, &requirement)
+        }
+    }
```
playfulFence: I would move these two to `src/config.rs`, under the `ActiveConfiguration` maybe? Seems like a better place to live for them. It also might already have some needed functionality implemented 🤔

src/tui.rs
```diff
@@ -56,7 +59,123 @@ impl Repository {
...
+                .cloned()
+                .unwrap_or_default()
+        };
+
+        for option_name in to_restore {
+            self.config.select(&option_name);
+        }
+    }
+
+    fn can_switch_method_option(&self, method_option: &str) -> bool {
```
playfulFence: Can also be moved to `config.rs` 🤔

src/tui.rs
```diff
@@ -586,11 +798,19 @@ impl App {
             .min(self.repository.current_level().len() - 1);
         let option = &self.repository.current_level()[selected];
 
+        let mut relationships = self.repository.config.collect_relationships(option);
+
+        if let GeneratorOptionItem::Option(option) = option {
+            if self.repository.can_switch_method_option(&option.name) {
+                relationships.disabled_by.clear();
```
playfulFence, line 772: Just a nitpick, but could you please leave a little comment on this? 😅 

Just to make sure future reader will understand it right away.

playfulFence (changes requested):

playfulFence (approved):
Okay, I think now I like how things look 👍🏼 

Thanks for putting up with my nitpicking, LGTM! 😄

KushalMeghani1644:
Thanks 😊

PR #315, opened 2026-04-02
Fix UI-breaking warnings

### LINKED ISSUE #308: Logged warnings break the UI completely
e.g. this

Image

makes the UI look like this

Image

playfulFence:
Closes https://github.com/esp-rs/esp-generate/issues/308.

I chose the option 2 from what @bjoernQ suggested - storing warnings into a buffer while TUI is active and flushing them after the terminal is restored.

https://github.com/user-attachments/assets/05df4d6b-6c93-4500-9715-aa6b33945552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix IDE settings and checks when using `-a/--name` (#302)
+- Fix UI-breaking warnings (#315)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


bjoernQ (approved):

MabezDev:
I still see a bit of breakage, albeit only on the terminal _once_ we've generated.

image

Its like we're not cleanly coming out of tui mode :thinking:

bjoernQ:
> I still see a bit of breakage, albeit only on the terminal once we've generated.

Seems to work fine for me on Windows 🤔

MabezDev:
might just be my terminal, but it didn't used to - probably nothing to worry about for now.

PR #325, opened 2026-04-20
Remove rust-analyzer.check.extraEnv

Bowen951209:
Remove `rust-analyzer.check.extraEnv` as it is redundant. `rust-analyzer.check.extraEnv` extends `rust-analyzer.cargo.extraEnv`, so defining it separately is unnecessary.

See also: <https://rust-analyzer.github.io/book/configuration.html#check.extraEnv>


SergioGasquez (approved):
Thanks for the simplification!

PR #326, opened 2026-04-21
Include esp-config.toml

### LINKED ISSUE #206: Put out config options in a seperate config.toml file
Context from here: https://github.com/esp-rs/esp-hal/issues/3020#issuecomment-3032177860, we should push users to keep their config separate from compilation options. 

After sleeping on it, it kinda sucks that we'll have to use an alias to include the config. Users can no longer run `cargo run`.  Maybe this should be a tui option instead? Or maybe we can include the config by default somehow.

SergioGasquez:
This PR isolates the runtime conifg generated by `esp-config` and the log levels in a `esp-config.toml` file.

This is not yet a multi-target solution, the multi-target solution would be something like (example supporting h2 and c6):

```toml
# .cargo/config.toml

[build]
target = "riscv32imac-unknown-none-elf"

[target.riscv32imac-unknown-none-elf]
# Required to obtain backtraces (e.g. when using the "esp-backtrace" crate.)
# NOTE: May negatively impact performance of produced code
rustflags = ["-C", "force-frame-pointers"]

[unstable]
build-std = ["core"]

[alias]
# ESP32-C6
build-c6 = [
  "build",
  "--target", "riscv32imac-unknown-none-elf",
  "--features", "chip-esp32c6",
  "--config", ".cargo/esp-config.toml",
  "--config", ".cargo/esp-config.esp32c6.toml",
]
run-c6 = [
  "run",
  "--target", "riscv32imac-unknown-none-elf",
  "--features", "chip-esp32c6",
  "--config", ".cargo/esp-config.toml",
  "--config", ".cargo/esp-config.esp32c6.toml",
]

# ESP32-H2
build-h2 = [
  "build",
  "--target", "riscv32imac-unknown-none-elf",
  "--no-default-features",
  "--features", "chip-esp32h2",
  "--config", ".cargo/esp-config.toml",
  "--config", ".cargo/esp-config.esp32h2.toml",
]
run-h2 = [
  "run",
  "--target", "riscv32imac-unknown-none-elf",
  "--no-default-features",
  "--features", "chip-esp32h2",
  "--config", ".cargo/esp-config.toml",
  "--config", ".cargo/esp-config.esp32h2.toml",
]
``` 

```toml
# .cargo/esp-config.esp32c6.toml
[env]
ESP_BOOTLOADER_ESP_IDF_CONFIG_ESP_IDF_VERSION = "0.2.0"

[target.riscv32imac-unknown-none-elf]
runner = "espflash flash --monitor --chip esp32c6"
``` 

```toml
# .cargo/esp-config.esp32h2.toml
[env]
ESP_BOOTLOADER_ESP_IDF_CONFIG_ESP_IDF_VERSION = "0.2.0"

[target.riscv32imac-unknown-none-elf]
runner = "espflash flash --monitor --chip esp32h2"
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [ opened, synchronize, reopened ]
     paths-ignore:
       - "**/README.md"
   merge_group:
@@ -22,7 +22,7 @@ env:
   CARGO_TARGET_DIR: ${{ github.workspace }}/target
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  MSRV: "1.86"
+  MSRV: "1.95"
   SSID: ""
   PASSWORD: ""
 
@@ -45,7 +45,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        chip: [esp32, esp32c2, esp32c3, esp32c6, esp32c5, esp32c61, esp32h2, esp32s2, esp32s3]
+        chip:
+          [
+            esp32,
+            esp32c2,
+            esp32c3,
+            esp32c6,
+            esp32c5,
+            esp32c61,
+            esp32h2,
+            esp32s2,
+            esp32s3,
+          ]
 
     steps:
       - uses: actions/checkout@v6
@@ -79,7 +90,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Generate and check project
-        run: cargo ${{ env.TOOLCHAIN }} xtask check ${{ matrix.chip }} ${{ fromJSON('["", "--all-combinations"]')[inputs.all_combinations || github.event_name == 'schedule'] }} ${{ fromJSON('["", "--build"]')[inputs.build || github.event_name == 'schedule'] }}
+        run: cargo ${{ env.TOOLCHAIN }} xtask check ${{ matrix.chip }} ${{
+          fromJSON('["", "--all-combinations"]')[inputs.all_combinations ||
+          github.event_name == 'schedule'] }} ${{ fromJSON('["",
+          "--build"]')[inputs.build || github.event_name == 'schedule'] }}
 
       - if: github.event_name == 'schedule'
         name: Run cargo-package
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 on:
   pull_request:
-    types: [ opened, synchronize, reopened ]
+    types: [opened, synchronize, reopened]
     paths-ignore:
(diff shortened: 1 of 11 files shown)


SergioGasquez:
Since this [requires at least 1.94](https://github.com/esp-rs/esp-generate/issues/206#issuecomment-4010924175), I guess we should wait until https://github.com/esp-rs/esp-hal/issues/5371 is merged/released and esp-hal requires a higher MSRV?

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ env:
   CARGO_TARGET_DIR: ${{ github.workspace }}/target
   CARGO_TERM_COLOR: always
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  MSRV: "1.86"
+  MSRV: "1.95"
```
bugadani: Raising MSRV and reformatting this file seems out of scope for this PR

src/cargo.rs
```diff
@@ -115,16 +115,14 @@ impl CargoToml {
                 Item::None => {
                     // alias = { package = "foo", version = "version" }
                     let update_renamed_dep = table.get_values().iter().find_map(|(k, p)| {
-                        if let Value::InlineTable(table) = p {
-                            if let Some(Value::String(name)) = &table.get("package") {
-                                if name.value() == package_name {
+                        if let Value::InlineTable(table) = p
+                            && let Some(Value::String(name)) = &table.get("package")
+                                && name.value() == package_name {
```
bugadani: Code formatting changes seem out of scope

bugadani:
We can land the content changes, but the rest is a bit annoying for #331 - I've done my best to avoid reformatting tui.rs to not cause unnecessary merge conflicts in that PR.

.github/workflows/ci.yml
SergioGasquez: I think the `include` of `.cargo/config.toml` was stabilized in 1.94, so I guess we needed to bump it, and went for 1.95 to match `esp-hal`

src/cargo.rs
SergioGasquez: The format CI check was failing after the msrv bump

.github/workflows/ci.yml
bugadani: The template's MSRV and esp-generate's MSRV are completely independent. Bumping esp-generate isn't really necessary in this PR.

bugadani (approved):

bugadani:
> This is not yet a multi-target solution

I don't think esp-generate will ever be able to handle multiple targets at once, considering they can have incompatible feature sets, and have incompatible pin- and memory layouts. What's the goal here?

SergioGasquez:
> I don't think esp-generate will ever be able to handle multiple targets at once, considering they can have incompatible feature sets, and have incompatible pin- and memory layouts. What's the goal here?

No precise goal, just wanted to experiment to see if this would open the doors for an easy way to handle multiple targets for some projects.

PR #327, opened 2026-04-21
Unselect options instead of preserving them

bugadani:
This PR replaces the logic to remember options with a visual indicator that lists what will be cleared.


bjoernQ (approved):
LGTM

PR #328, opened 2026-04-21
Move chip selection to the TUI

bugadani:
This PR reworks the chip selection process:

- `--chip` can still specify an initial selection, but
- The TUI has a chip row, that the user can use to select a chip
- Selecting a chip essentially reloads the ui, while keeping compatible options selected
- Template-wise, the compatibility check that may hide or show a config item can now use arbitrary config groups (TODO: rework flashing and async options)

The idea is that future external templates may not even have a chip option, so there is no point in having _special cased_ handling for them.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Generate a sample project
         run: |
-          cargo run --release --quiet --no-default-features -- --headless --chip=esp32c6 --output-path=. test-project
+          cargo run --release --quiet --no-default-features -- --headless --output-path=. -o esp32c6 test-project
 
       - name: Check links in generated project
         uses: lycheeverse/lychee-action@v2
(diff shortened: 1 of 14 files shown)


bjoernQ (approved):
this is really nice usability wise

SergioGasquez (approved):
LGTM! Thanks for the improvement, I did some testing, and the only small and ignorable nitpick is that I found a bit confusing the "will disables: esp32" message in: 

Image

The fact that esp32 is the default and the rest show that it will disable it, can be a bit confusing for some users. Maybe for chip selection, I wouldnt show the "will disable..."

bugadani:
What about "will deselect"? I'm completely de-specialising `chip` so we won't really know when to print something and when not to.

SergioGasquez:
> What about "will deselect"? I'm completely de-specialising `chip` so we won't really know when to print something and when not to.

Yeah, that would be a better wording imho.

MabezDev (approved):

PR #330, opened 2026-04-22
Move data from binary to template

bugadani:
This PR moves chips, wokwi board names and the module list into the template. To achieve this, the following features have been added:

- `sets` directives: allow selecting an option to update a template variable with a string (e.g. wokwi board name) or a list of strings (e.g. pin limitation lists)
- `!Include` an external yaml file as an option category. This lets us separate the long lists into their own files

The binary still makes a bunch of data available for the templates, that may count as hard-coded/specialized for the current bundled template:
 - metadata:
   - reclaimed memory sizes - it's less error-prone to provide than it is to copy
   - the pin lists/limitations
 - the binary expands the list of pins-to-remove into code, this is hard to formulate in a yaml template. As a solution, we could use some list iteration feature, but we probably should use a different templating engine if we wanted to resolve this


MabezDev (approved):
LGTM! I suppose at the end of this we'll have our own little template SDK :D.

PR #332, opened 2026-04-22
Fix neovim

### LINKED ISSUE #324: Neovim Config Template Doesn't Work
# Problem

rust-analyzer fails in Neovim Xtensa generated config, with the following error:
```
error: 'rust-analyzer' is not installed for the toolchain 'esp'
```
This suggests that **the stable toolchain is not being correctly selected.**

The template is intended to specify the stable toolchain. However, it appears that the `rust-analyzer.server` option only exists in the VSCode rust-analyzer extension:
https://github.com/esp-rs/esp-generate/blob/0f0de2543376654ca172593415ede3145631e9c5/template/.nvim.lua#L15 
, because it is not listed in [rust-analyzer book](https://rust-analyzer.github.io/book/configuration.html).

Another problem is this typo,

```diff
 -- template/.nvim.lua

 --INCLUDEFILE option("neovim")
 -- You must enable the exrc setting in neovim for this config file to be used.
 local rust_analyzer = {
     cargo = {
         --REPLACE riscv32imac-unknown-none-elf rust_target
         target = "riscv32imac-unknown-none-elf",
         allTargets = false,
     },
 }
 --IF option("xtensa")
 --REPLACE esp rust_toolchain
-rust_analyzer.cargo.extraEnv = { RUST_TOOLCHAIN = "esp" }
+rust_analyzer.cargo.extraEnv = { RUSTUP_TOOLCHAIN = "esp" }
 --REPLACE esp rust_toolchain
-rust_analyzer.check = { extraEnv = { RUST_TOOLCHAIN = "esp" } }
-rust_analyzer.server = { extraEnv = { RUST_TOOLCHAIN = "stable" } }
+rust_analyzer.check = { extraEnv = { RUSTUP_TOOLCHAIN = "esp" } }
+rust_analyzer.server = { extraEnv = { RUSTUP_TOOLCHAIN = "stable" } }
 --ENDIF
 
 -- Note the neovim name of the language server is rust_analyzer with an underscore.
 vim.lsp.config("rust_analyzer", {
     settings = {
         ["rust-analyzer"] = rust_analyzer
     },
 })
```

# Solution

I was able to get rust-analyzer working with the following configuration:

```lua
local rust_analyzer = {
	cargo = {
		target = "xtensa-esp32-none-elf",
		allTargets = false,
	},
}
rust_analyzer.cargo.extraEnv = { RUSTUP_TOOLCHAIN = "esp" }

vim.lsp.config("rust-analyzer", {
	cmd = function()
		return { "rustup", "run", "stable", "rust-analyzer" }
	end,
	settings = {
		["rust-analyzer"] = rust_analyzer,
	},
})

```
I am using `rustaceanvim`. According to [its manual](https://github.com/mrcjkb/rustaceanvim/blob/824491c58bdba1f6f5e35e64e0ab8b2e2425d6b2/doc/rustaceanvim.txt#L173), the correct key is `rust-analyzer`.  However, in the template:
https://github.com/esp-rs/esp-generate/blob/0f0de2543376654ca172593415ede3145631e9c5/template/.nvim.lua#L18
, so we should remember to handle both if we are implementing the following:

# Suggestion

Add two sub-options for Neovim config generation so users can choose between manual LSP setup and rustaceanvim. Then 　`esp-generate` can generate the appropriate key accordingly:

| LSP setup  | key name  |  
|---|---|
|  manual | `rust_analyzer`  | 
|  rustaceanvim | `rust-analyzer`  |

bugadani:
Closes https://github.com/esp-rs/esp-generate/issues/324

cc @Bowen951209 if you could please test this

### DIFF
--- a/template/.nvim.lua
+++ b/template/.nvim.lua
@@ -13,8 +13,12 @@ rust_analyzer.cargo.extraEnv = { RUSTUP_TOOLCHAIN = "esp" }
 rust_analyzer.server = { extraEnv = { RUSTUP_TOOLCHAIN = "stable" } }
 --ENDIF
 
+--IF option("neovim-rustaceanvim")
+vim.lsp.config("rust-analyzer", {
+--ELSE
 -- Note the neovim name of the language server is rust_analyzer with an underscore.
 vim.lsp.config("rust_analyzer", {
+--ENDIF
     settings = {
         ["rust-analyzer"] = rust_analyzer,
     },
--- a/template/.nvim.lua
+++ b/template/.nvim.lua
@@ -9,8 +9,8 @@ local rust_analyzer = {
 }
 --IF option("xtensa")
 --REPLACE esp rust_toolchain
-rust_analyzer.cargo.extraEnv = { RUST_TOOLCHAIN = "esp" }
-rust_analyzer.server = { extraEnv = { RUST_TOOLCHAIN = "stable" } }
+rust_analyzer.cargo.extraEnv = { RUSTUP_TOOLCHAIN = "esp" }
+rust_analyzer.server = { extraEnv = { RUSTUP_TOOLCHAIN = "stable" } }
 --ENDIF
 
 -- Note the neovim name of the language server is rust_analyzer with an underscore.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed neovim config (#332)
+
 ### Removed
 
 ## [1.3.0-rc.0] - 2026-04-17
--- a/template/template.yaml
+++ b/template/template.yaml
@@ -190,8 +190,14 @@ options:
         display_name: Add settings for Helix Editor
 
       - !Option
-        name: neovim
+        name: neovim-classic
+        selection_group: neovim
         display_name: Add settings for Neovim
+
+      - !Option
+        name: neovim-rustaceanvim
+        selection_group: neovim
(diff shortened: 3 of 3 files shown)


Bowen951209:
Thanks! I'll test it later this day.

template/.nvim.lua
```diff
@@ -9,12 +9,16 @@ local rust_analyzer = {
 }
 --IF option("xtensa")
 --REPLACE esp rust_toolchain
-rust_analyzer.cargo.extraEnv = { RUST_TOOLCHAIN = "esp" }
-rust_analyzer.server = { extraEnv = { RUST_TOOLCHAIN = "stable" } }
+rust_analyzer.cargo.extraEnv = { RUSTUP_TOOLCHAIN = "esp" }
+rust_analyzer.server = { extraEnv = { RUSTUP_TOOLCHAIN = "stable" } }
```
Bowen951209, line 13: `rust_analyzer.server` is not a valid option for the standard rust-analyzer used in Neovim. As mentioned in #324, this setting appears to be exclusive to the VSCode extension and has no effect in this setup.

The solution I found in #324 is to set the `cmd` field instead:

```diff 
vim.lsp.config("rust-analyzer", {
+    cmd = { "rustup", "run", "stable", "rust-analyzer" },
    settings = {
        ["rust-analyzer"] = rust_analyzer,
    },
})
```

Bowen951209 (changes requested):

template/.nvim.lua
bugadani: Perhaps it would be best if you sent in the fix as a pull request, since you are the one who can make sure it is actually working.

Bowen951209: Sure. I'll do it.

PR #335, opened 2026-04-28
Add coding agent guidance file

### LINKED ISSUE #319: Optionally include an AGENTS.md file
To help users which are using coding agents we could provide an AGENTS.md file

see https://agents.md/

Example: https://github.com/openai/codex/blob/main/AGENTS.md

I don't have a lot of ideas for what to include right now but at least we could tell the agent where to look for docs etc.

Another example (but that's a skill): https://github.com/google-gemini/gemini-skills/blob/main/skills/gemini-api-dev/SKILL.md

SergioGasquez:
Any hints on how to improve the content of the file is more than welcome!


bjoernQ:
Not sure if it would actually help or confuse the coding agents but

- the critical rules where they say that training data is outdated ( like https://github.com/google-gemini/gemini-skills/blob/main/skills/gemini-api-dev/SKILL.md ) 
- where to find the correct documentation (same document)

otherwise a great start - I guess this is something we'll see a few contributions for over time

template/CODING_AGENT_GUIDANCE.md
```diff
@@ -0,0 +1,45 @@
+#INCLUDEFILE option("coding-agent-guidance")
+#INCLUDE_AS {coding-agent-guidance-file}
+#REPLACE {coding-agent-guidance-file} coding-agent-guidance-file
+# {coding-agent-guidance-file}
+
+This project was generated by `esp-generate` for Espressif MCUs (`no_std`).
+
+## Quick orientation
+
+#REPLACE esp32c6 mcu
```
MabezDev, line 10: Imo the quick orientation and common tasks are noise.

Agents are pretty capable these days, and know about rust, how projects are structured etc - what we need to explain (briefly) are the weird bits that we do differently, or constraints for our environment, things like:

- config stuff (where to find them, briefly how it works (via env vars)
- no_std (already mentioned below, nice!), plus alloc
- flashing and tools:  to use (either espflash, probe-rs) - if we get this right, we can stop a lot of braindead issues coming are way - think about what we can suggest, for example if probe-rs isn't working, we can suggest checking the correct port is being used to the agent which will hand hold the user for us.

Probably a bunch more things, but in general I believe file should be minimal by design, as noise is worse than silence - only include things we KNOW will be helpful that an agent CANNOT know before seeing the project.

MabezDev (approved):
Better, thanks!

PR #336, opened 2026-04-29
Fix Neovim template

### LINKED ISSUE #324: Neovim Config Template Doesn't Work

Bowen951209:
**Fix how the stable toolchain is specified in Neovim (Fix #324)**

This PR is a follow-up to #332.
It adds support for both classic Neovim LSP and [rustaceanvim](https://github.com/mrcjkb/rustaceanvim) configurations configurations.

### DIFF
--- a/template/template.yaml
+++ b/template/template.yaml
@@ -190,8 +190,14 @@ options:
         display_name: Add settings for Helix Editor
 
       - !Option
-        name: neovim
+        name: neovim-classic
+        selection_group: neovim
         display_name: Add settings for Neovim
+
+      - !Option
+        name: neovim-rustaceanvim
+        selection_group: neovim
+        display_name: Add settings for Neovim (with rustaceanvim)
 
       - !Option
         name: vscode
--- a/template/template.yaml
+++ b/template/template.yaml
@@ -191,12 +191,10 @@ options:
 
       - !Option
         name: neovim-classic
-        selection_group: neovim
         display_name: Add settings for Neovim
 
       - !Option
         name: neovim-rustaceanvim
-        selection_group: neovim
         display_name: Add settings for Neovim (with rustaceanvim)
 
       - !Option
--- a/template/template.yaml
+++ b/template/template.yaml
@@ -191,10 +191,12 @@ options:
 
       - !Option
         name: neovim-classic
+        selection_group: neovim
         display_name: Add settings for Neovim
 
       - !Option
         name: neovim-rustaceanvim
+        selection_group: neovim
         display_name: Add settings for Neovim (with rustaceanvim)
 
       - !Option
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed neovim config (#332)
+
 ### Removed
 
 ## [1.3.0-rc.0] - 2026-04-17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed neovim config (#332)
+- Fixed neovim config (#336)
 
 ### Removed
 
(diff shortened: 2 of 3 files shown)


template/template.yaml
```diff
@@ -190,9 +190,13 @@ options:
         display_name: Add settings for Helix Editor
 
       - !Option
-        name: neovim
+        name: neovim-classic
         display_name: Add settings for Neovim
 
+      - !Option
+        name: neovim-rustaceanvim
+        display_name: Add settings for Neovim (with rustaceanvim)
```
bugadani, line 200: What's the point of making these coexist? Will this not confuse users?

Bowen951209: I think some users may want to include editor config in version control so others can clone the repo and start working immediately in their editor.

bugadani: To me it sounds unlikely. Also, sharing editor config like that is a bit smelly.

Anyway, these options should be excluded from the CI-checked configurations, we're blowing up CI time with something that doesn't have any observable effect.

Bowen951209: Got it. I’ll revert the change to disallow coexistence.

bugadani (approved):

PR #339, opened 2026-04-29
Remove special-casing chip, vol. 2

bugadani:
This PR makes it possible to create templates without a `chip` group. This is done by detecting the option group and only injecting some variables when it is present.

Some of this special casing is needed to not duplicate data from E-M-G. The hard-coded pin pluck codegen is necessary because the template language doesn't support a for-each option, but we also don't have the capability to resolve the chip limitation categories into actual chips purely in the template. I think this level of special-casing is acceptable, although a for-each syntax to iterate `sets: key: [list of values]` would help external templates.

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -61,7 +61,6 @@ enum CheckResult {
 }
 
 pub fn check(
-    project_path: &Path,
     chip: Chip,
     probe_rs_required: bool,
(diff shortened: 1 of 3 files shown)


MabezDev (approved):
Thanks!

PR #352, opened 2026-07-27
Separate the SDK

Hahihula:
Carves the pure, IO-free template contract out of the binary into a new crate. First step toward external templates.


esp-template-sdk/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name         = "esp-template-sdk"
+version      = "0.1.0"
+edition      = "2024"
+rust-version = "1.86"
+description  = "Versioned template contract for esp-generate: option-tree parsing, the relationship engine, and the pure file-directive processor."
+repository   = "https://github.com/esp-rs/esp-generate"
+license      = "MIT OR Apache-2.0"
+publish      = false
```
MabezDev: Is this temporary, to avoid publishing before we're ready?

Cargo.toml
```diff
@@ -23,23 +36,24 @@ default = ["update-informer"]
 update-informer = ["dep:update-informer"]
 
 [dependencies]
-anyhow          = "1.0.100"
+esp-template-sdk = { path = "esp-template-sdk" }
```
MabezDev: We should also set a version here.

esp-template-sdk/src/process.rs
```diff
@@ -0,0 +1,990 @@
...
+                            .get(var_name)
+                            .map(|value| (pattern, value.to_string()))
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<_>>();
+
+            if !replacements.is_empty() {
+                replace = Some(replacements);
```
MabezDev: #REPLACE appears to remain pending inside an inactive #IF branch. Since it is only cleared after a line is emitted, it can affect the first active line after #ENDIF. Could we prevent that and add a regression test for this case?

src/main.rs
```diff
@@ -1123,6 +929,7 @@ fn process_options(template: &Template, args: &Args) -> Result<()> {
         selected,
         flat_options,
         options: template.options.clone(),
+        facts: None,
```
MabezDev: Could we wire chip facts into the production configurations in this PR? Both this headless configuration and the TUI initialize `facts` as `None`. That leaves every `requires_capabilities` declaration unconstrained even though the SDK unit test passes. The facts need to be installed before validation and refreshed whenever the selected chip changes.

esp-template-sdk/src/process.rs
```diff
@@ -0,0 +1,990 @@
...
+            {
+                include_file = eval_bool(&mut engine, cond, line_no, "#INCLUDEFILE")?;
+                continue;
+            } else if let Some(include_as) = trimmed
+                .strip_prefix("//INCLUDE_AS ")
+                .or_else(|| trimmed.strip_prefix("#INCLUDE_AS "))
+                .or_else(|| trimmed.strip_prefix("--INCLUDE_AS "))
+            {
+                let mut include_as = include_as.trim().to_string();
+                for (key, value) in &facts.values {
```
MabezDev: Could we make this a single-pass? `facts.values` is a `HashMap`, so repeated global replacements have randomized order. If one value contains another `{name}` placeholder, the resolved path can depend on map iteration order. Parsing placeholders from the original path and substituting each token once would keep generation deterministic.

esp-template-sdk/src/contract.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+impl std::error::Error for ParseVersionError {}
+
+impl FromStr for Version {
+    type Err = ParseVersionError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let core = s.split(['-', '+']).next().unwrap_or(s);
```
MabezDev: The template gate is specified as semver, but stripping the suffix makes `2.0.0-rc.1` compare equal to `2.0.0`. That can let a prerelease generator satisfy a final-release floor. Could we use the `semver` crate for contract versions and keep any lenient host-tool parsing separate?

esp-template-sdk/src/process.rs
```diff
@@ -0,0 +1,990 @@
...
+/// directory. Malformed directives are reported as [`ProcessError`] with the
+/// offending line, never panicked.
+///
+/// A key absent from [`Facts::values`] is intentionally left unsubstituted
+/// (the literal placeholder survives) — that's the template's "override or
+/// keep the default" idiom, e.g. `rust-toolchain.toml` keeps `channel = "stable"`
+/// unless an explicit `rust_toolchain` was chosen. Distinguishing an optionally-
+/// absent value from a typo needs the full declared-key universe, which only
+/// the binary's `check` command has; flagging unknown keys is `check`'s job.
+pub fn process_file(
```
MabezDev, line 561: Nit: To me, a `fn` called `process_file`, I'd expect to provide a path to the file, not the contents.

esp-template-sdk/src/process.rs
```diff
@@ -0,0 +1,990 @@
...
+
+    engine.add_function("option", move |cond: &str| -> bool {
+        selected.iter().any(|c| c == cond)
+    });
+
+    engine.add_function("group_selected", move |group: &str| -> bool {
+        selected_groups.iter().any(|g| g == group)
+    });
+
+    engine.add_function("chip_has", move |symbol: &str| -> bool {
```
MabezDev: How do we want to preserve the hard-error guarantee for unknown capability names here? With only the selected chip’s symbol set, a typo is indistinguishable from a valid capability that this chip does not have, so both silently return `false`. If the full validation is landing with `check`, I think this seam still needs access to the per-spec known-symbol vocabulary before we call the v1 surface complete.

esp-template-sdk/src/process.rs
MabezDev, line 561: I'm also not sure if we have to worry about non-UTF8 bytes, but it might be something we have to deal with so we should think about it now.

MabezDev (commented):
First round of review from my side, looking good so far! 

Just some tweaking to do I think :)

esp-template-sdk/Cargo.toml
Hahihula: I just thought that before we have the ability to actually use external templates, there is no use for the crate outside of the repository.

esp-template-sdk/src/process.rs
Hahihula: The name of the function was just kept as it was, basically moved from main.rs. Would `process_file_content` sound better? WDYT?

Hahihula: I was thinking the `contents: &str` guarantees valid UTF-8, am I missing something?

esp-template-sdk/src/process.rs
Hahihula: Thanks for the catch. I completely missed the case of replacement of replacement.

esp-template-sdk/src/process.rs
Hahihula: The full AST validation should land in the check. But I will add check agains union all known symbol vocabulary.

esp-template-sdk/src/contract.rs
Hahihula: You are right this approach is faulty.

src/main.rs
Hahihula: I was thinking this is more part of the second step but you are right. this way it would be kept unconstrained.

esp-template-sdk/src/process.rs
Hahihula: I bound the REPLACE just to next line. that should solve the issue but i am not entirely sure it's the correct solution, but we probably will not need multiline REPLACE...

esp-template-sdk/src/contract.rs
```diff
@@ -0,0 +1,327 @@
...
+/// Host *tool* versions (rustc, espflash, probe-rs) are a different problem:
+/// they are reported in loose formats and their prereleases should be treated
+/// as good enough. That leniency lives in the binary's `check` module and is
+/// deliberately not shared with this type.
+pub use semver::Version;
+
+/// The `spec_version` this SDK implements. A breaking contract change bumps
+/// this; additive growth does not (it raises a template's
+/// [`min_generator_version`] floor instead).
+pub const SPEC_VERSION: u32 = 1;
```
bugadani: This is unnecessary, we can use the crate's version as the spec version.

esp-template-sdk/src/contract.rs
```diff
@@ -0,0 +1,327 @@
...
+        name: "rust_toolchain",
+        kind: FeatureKind::Value,
+        since: V1_BASELINE,
+    },
+    Feature {
+        name: "reserved_gpio_code",
+        kind: FeatureKind::Value,
+        since: V1_BASELINE,
+    },
+    // Directives (see `process::process_file`).
```
bugadani: These will be replaced by https://crates.io/crates/somni-template soon enough. I suspect adding new template features will require bumping SDK version by one minor version, but I don't see how we would version it any further.

Even `INCLUDEFILE` and `INCLUDE_AS` could in theory be emulated with conditional includes.

The point is, we maintain the template engine, we control its capabilities, we shouldn't need to maintain a second one for esp-generate.

esp-template-sdk/src/contract.rs
```diff
@@ -0,0 +1,327 @@
...
+pub use semver::Version;
+
+/// The `spec_version` this SDK implements. A breaking contract change bumps
+/// this; additive growth does not (it raises a template's
+/// [`min_generator_version`] floor instead).
+pub const SPEC_VERSION: u32 = 1;
+
+/// The first esp-generate release that exposes the `spec_version` 1 fact API.
+/// Every baseline v1 feature is tagged with this.
+pub const V1_BASELINE: Version = release(2, 0, 0);
```
bugadani: This is a weird concept - we usually only bump version before releasing something, so we wouldn't be able to use new SDK features immediately.

esp-template-sdk/src/process.rs
bugadani: `&str` must be valid UTF-8, yes

bugadani (commented):
Actually, what's the point of all this? To write a template, you don't need a rust crate - it's a folder/zip/whatever we decide containing only template files, and a yaml descriptor. The only active part here is esp-generate, there's nothing else reading/rendering templates. So why are we introducing ~1800 lines of code?

esp-template-sdk/src/contract.rs
bugadani: So, in my head, we can get away with only the SDK version, which is also the SDK crate version:

- What template can be loaded is determined by what SDKs are built into esp-generate
- What a template can do is determined by the SDK version it references
- We'll adhere to semver, so a template that declares SDK version x.y, will be compatible with SDK version x.y+n.

Based on this, do we still need this baseline version?

esp-template-sdk/src/process.rs
MabezDev: somni will likely take over these parsing steps, so I think we can roll with this for now and swap it out later.

esp-template-sdk/src/process.rs
MabezDev: ah if this is just how it was, we can leave it for now.

MabezDev: I was more wondering, should we take `&[u8]` instead, incase we need to process non utf8 chars or symbols parsed by users, but I think we can roll with `&str` for now, and we can change it later if we need it.

bugadani: What symbols do you have in mind?

MabezDev: It's not so much about the symbols I have in mind. It's that I know how our config system uses environment variables and environment variables are essentially OS strings, not UTF-8 strings - I think we can just change this when/if we want to though to remove such a limitation.

esp-template-sdk/src/contract.rs
Hahihula: It sounds right; I actually can not thing of a counterargument.

MabezDev: Agreed, we can drop the baseline record in favour using semver for the sdk version.

Hahihula:
The manifest in the next phase should remove the need for 'includefile' altogether.

MabezDev (approved):
This is looking great! I took it for a spin and everything seems to be in order. Let's move on to the next stage.

PR #357, opened 2026-08-07
migrating the binary to start using the SDK and manifest file in the template

Hahihula:
After this, adding the check() and then actually enabling external templates remains.


src/builtin.rs
```diff
@@ -0,0 +1,135 @@
+//! Binary-provided option-tree fragments — the `builtin:` include scheme.
```
bugadani: I don't mind the `builtin:` scheme, but this must actually be part of the SDK, otherwise we'll end up with templates that behave differently under different binaries. In essence, a template will declare dependency on a specific SDK, and it will have a different, hidden dependency on "something else".

Hahihula: You are right that it should be part of the SDK. I moved it there.

esp-template-sdk/src/builtin.rs
```diff
@@ -0,0 +1,242 @@
...
+                "  - !Option\n    \
+                   name: {}\n    \
+                   display_name: {}\n    \
+                   selection_group: chip\n",
+                chip.name, chip.display_name
+            ));
+
+            if let Some(board) = &chip.wokwi_board {
+                yaml.push_str(&format!("    sets:\n      wokwi_board: {board}\n"));
+            }
```
bugadani: I think this is supposed to be `wokwi-board`, but we also don't need it anymore, the template can just substitute the right variable from the chip struct/based on the name of the chip.

esp-template-sdk/src/process.rs
```diff
@@ -217,20 +217,36 @@ fn is_somni_identifier(name: &str) -> bool {
...
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ProcessError {
-    /// 1-based line number in the source template file.
-    pub line: usize,
+    /// 1-based line in the template file, when the failure has one.
+    ///
+    /// `None` for failures that are not *in* a file — a manifest condition, for
+    /// instance, which is authored in `metadata.toml` and has no line the SDK
+    /// could name. Modelled as an option rather than a sentinel because a
+    /// rendered `line 0:` is worse than no line at all.
```
bugadani: No need to explain design decisions, especially when they align with Rust conventions. On the other hand, if this is 1-based, it should be `Option<NonZeroUsize>` probably

esp-template-sdk/src/process.rs
```diff
@@ -538,157 +534,121 @@ fn strip_keyword<'a>(rest: &'a str, keyword: &str) -> Option<&'a str> {
...
-    line: usize,
-) -> Result<bool, ProcessError> {
-    let marker = style.marker();
-    let probe = format!("{marker}if {cond}\nyes\n{marker}endif\n");
-
-    let render = Template::compile(&probe, syntax).and_then(|t| t.render(build_env(shared)));
-
-    if let Some(reason) = shared.unknown.borrow_mut().take() {
-        return Err(ProcessError {
+impl Renderer {
```
bugadani: docstrings on an impl block is pretty unusual and I'm not sure if rustdoc even generates the documentation for it - I think a single impl block for Renderer would be nicer, and you should move the doc comment on the method where it belongs.

src/toolchain.rs
```diff
@@ -7,19 +7,17 @@ use std::thread;
...
-/// and then derive per-chip filtered views via [`toolchains_for_chip`] without
-/// spawning any more subprocesses.
+/// Scanning is expensive — one `rustc +tc --print target-list` per toolchain —
+/// so we capture it once up front and derive per-chip filtered views via
+/// [`toolchains_for_chip`] without spawning any more subprocesses.
 #[derive(Debug, Clone)]
 pub struct ToolchainInfo {
     pub name: String,
     pub targets: BTreeSet<String>,
-    pub version: Option<check::Version>,
```
bugadani, line 22: This removal seems out of scope for this PR.

template/metadata.toml
```diff
@@ -0,0 +1,78 @@
...
+# source path. Anything not listed here is emitted as-is; anything under
+# `.template/` is never emitted at all, which is where partials and the option
+# tree's `!Include` fragments live.
+#
+# `when` is a somni expression evaluated against the same facts a template body
+# sees, so `option("wokwi")` means exactly what it means inside a file.
+
+[[emit]]
+path = "wokwi.toml"
+when = 'option("wokwi")'
```
bugadani: This is really verbose and that makes it hard to scan through. I think I prefer the old `%includefile` directive on the files directly, otherwise I think this should be reformatted as an inline list:

```toml
emit = [
    { path = "wokwi.toml",   when = 'option("wokwi")' },
    { path = "diagram.json", when = 'option("wokwi")' },
    # ...
]
```

CHANGELOG.md
```diff
@@ -33,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Removed the BLE `bleps` option. (#341)
+- The `#%includefile` and `#%include_as` directives. Which files a generated project gets, and what they are called, is the manifest's job now — a file's own body no longer decides whether that file exists. (#357)
```
bugadani, line 39: You don't have to write changelog for the yet-to-be-released SDK, and template changes aren't user-visible before the split is released.

Hahihula: ok ... the pipeline changelog check was screaming :-)

esp-template-sdk/src/process.rs
Hahihula: sorry, the docstring remained from the original function, will fix it...

esp-template-sdk/src/process.rs
Hahihula: I actually was not aware of NonZeroUsize existance, thanks

src/toolchain.rs
Hahihula: ok, i keep it here and add the clippy #[allow(dead_code)]

bugadani: why? what did you change that no longer needs the toolchain version to be stored in memory, and how does that relate to the intent of this PR?

Hahihula: The check check msrv there is no need for version here. The version costs one extra subprocess per installed toolchain. The picker offers stable/nightly.

bugadani: What I read out of the current version of the code is: we now allow picking an incompatible (old) toolchain, and fail when the user is done with picking settings. That is not desireable UX, we shouldn't offer options that we are going to reject later, discarding user input in the process.

bugadani: _this_ is what I mean by this is out of scope for the PR, not the field in the struct.

Hahihula: ok, I finally got it... Sorry, you are right

esp-template-sdk/src/process.rs
```diff
@@ -450,245 +393,167 @@ pub fn is_safe_relative_path(path: &str) -> bool {
...
+    /// path is not code, and a literal lookup lets substitution-only names
+    /// contain dashes. An unknown `{name}` survives verbatim, which is the
+    /// override-or-default idiom; telling an optional absence from a typo needs
+    /// the full declared-key universe, so that is `check`'s job.
+    pub fn output_path(&self, path: &str) -> String {
+        interpolate(path, &self.shared.facts.values)
+    }
+
+    /// Evaluate a condition by rendering it as a one-line template, so
+    /// conditions and template bodies share one engine and one [`Env`] — there
```
bugadani: What's going on here, what prevents us from straight evaluating a `somni-expr` expression? Building a template just to do the same thing is a bit excessive.

template/.vscode/settings.json
```diff
@@ -1,4 +1,3 @@
-//%includefile option("vscode")
 {
   "rust-analyzer.cargo.allTargets": false,
   //REPLACE riscv32imac-unknown-none-elf rust_target
```
MabezDev: These two `//REPLACE` lines are from the old directive syntax, so they now get copied through verbatim. Generating for an ESP32-S3 produces a settings.json that still says `"rust-analyzer.cargo.target": "riscv32imac-unknown-none-elf"`, so these want `{{ chip.rust_target }}` and `{{ rust_toolchain }}` instead.

src/main.rs
```diff
@@ -26,32 +26,83 @@ use std::{
...
+/// Read before anything else touches the template, so an incompatible template
+/// fails with one line rather than an unknown-fact error part-way through
+/// generation.
+static MANIFEST: LazyLock<manifest::Manifest> = LazyLock::new(|| {
+    let source = SOURCE
+        .get(manifest::MANIFEST_PATH)
+        .unwrap_or_else(|| panic!("bundled template missing `{}`", manifest::MANIFEST_PATH));
+
+    let manifest =
+        manifest::Manifest::parse(source).expect("invalid manifest in the bundled template");
```
MabezDev: The manifest docs promise an incompatible template fails with one clear line, but this `expect` panics instead. With external templates next, a parse failure or a version mismatch should come back as an error from `main`, not a panic.

src/main.rs
```diff
@@ -823,33 +869,38 @@ fn main() -> Result<()> {
...
-            &mut load_partial,
-        )
-        .map_err(|e| anyhow::anyhow!("{source_path}:{e}"))?;
-        let Some(processed) = processed else {
-            continue; // excluded by `includefile`
+    for (source_path, contents) in SOURCE.files() {
+        let manifest::Emit::When { condition, output } = MANIFEST.emit(source_path) else {
+            continue;
         };
+        if let Some(condition) = &condition {
```
MabezDev: `when` is evaluated inside the write loop, so a bad condition aborts after `fs::create_dir` with earlier files already written. Evaluating every rule's condition once before the first write would keep that failure clean.

src/main.rs
```diff
@@ -823,33 +869,38 @@ fn main() -> Result<()> {
...
+            }
+        }
+
+        let processed = renderer
+            .render(contents, &mut load_partial)
+            .map_err(|e| anyhow::anyhow!("{source_path}:{e}"))?;
 
-        // Reject any output path that escapes the project dir, even after
-        // the SDK already validates `#INCLUDE_AS` paths.
+        let out_path = match &output {
```
MabezDev: `output_path` leaves an unknown `{name}` verbatim (the `src/{nope}.rs` case in the process tests pins this), so a missing fact silently writes a file literally named `{coding_agent_guidance_file}`. An unresolved placeholder in an `as` path should fail here instead.

src/manifest.rs
```diff
@@ -0,0 +1,306 @@
...
+                );
+            }
+        }
+
+        Ok(manifest)
+    }
+
+    /// What to do with the template file at `path`.
+    pub fn emit(&self, path: &str) -> Emit {
+        if path == MANIFEST_PATH || path == OPTION_TREE_PATH || path.starts_with(RESERVED_DIR) {
```
MabezDev: An `emit` rule for anything under `.template/`, or for `metadata.toml` / `template.yaml`, is never consulted because of this early return. That is the same silently inert rule `validate_paths` exists to catch, so it is worth rejecting at parse time.

src/lib.rs
```diff
@@ -68,6 +71,41 @@ impl Chip {
...
+            Chip::Esp32s3 => "ESP32-S3",
+        }
+    }
+
+    /// The Wokwi board this chip simulates as, if Wokwi supports it.
+    ///
+    /// Wokwi support is not a chip *capability* — it's a property of an
+    /// external simulator — so it can't come from `esp-metadata`. It rides
+    /// along with the chip because it is keyed by chip and nothing else.
+    pub fn wokwi_board(self) -> Option<&'static str> {
```
MabezDev, line 97: `wokwi_board`, `rust_target` and `dram2_uninit_size` are binary-supplied fields templates interpolate, and none of them are registered in `contract`, so `sdk_version` does not cover them. Same hidden dependency that pushed `builtin:chip` into the SDK, relocated to the chip struct.

MabezDev (changes requested):

template/.vscode/settings.json
bugadani:
```
//% replace "riscv32imac-unknown-none-elf" with rust_target`
...
//% endreplace
```

 should also work

esp-template-sdk/src/process.rs
Hahihula: should be now properly delegated to the somni-expr

template/.vscode/settings.json
Hahihula: hmm :-( my testing script comparing generation before PR and after PR was silently ignoring dotfiles and i was overconfident it worked :-(
Should be fixed now

src/main.rs
Hahihula: it should now fail with clean message

src/main.rs
Hahihula: unresolved placeholder now fails

src/manifest.rs
Hahihula: Manifest::parse now rejects a rule for .template/…, metadata.toml or template.yaml

src/lib.rs
Hahihula: I now moved them to SDK as ChipFacts.

esp-template-sdk/src/process.rs
```diff
@@ -169,46 +125,88 @@ pub struct Vocabulary {
...
-    /// The map must hold **every** field the template could reference, not just
-    /// the ones true for this chip: somni reports an unknown field as an error,
-    /// so a capability the chip merely lacks has to be present-and-`false`, or
-    /// a cross-chip `#%if chip.soc_has_wifi` would fail instead of taking its
-    /// else branch. Getting that union right is what buys the typo diagnostic —
-    /// `chip.soc_has_wfi` is then the *only* unknown name.
-    pub chip: Option<IndexMap<String, FactValue>>,
+    /// A scalar wins a name clash, so a metadata symbol called `rust_target`
+    /// could never displace the field this SDK promises.
+    fn fields(&self) -> IndexMap<Box<str>, FactValue> {
```
bugadani: This is really inefficient:
- ChipFacts already allocates a few hundered times when you call it
- This funciton then clones basically everything into an IndexMap of `Box<str> -> FactValue`
- `chip_symbols` also clones everything into an IndexMap of `String -> FactValue` (why the different types)

Chip data is static, we should be computing these maps once and only once. Instead, we recreate the whole thing for every app frame we render.

esp-template-sdk/src/process.rs
```diff
@@ -375,35 +402,56 @@ fn build_env(shared: &Rc<Shared>) -> Env {
...
-    env.value("has_reserved_pins", shared.facts.has_reserved_pins);
+    sink.fact(
+        "has_reserved_pins",
+        TypedValue::Bool(shared.facts.has_reserved_pins),
+    );
+
+    // `Env::new` pre-registers a `str` conversion but a bare `Context` does not,
+    // and the upstream one is private. Registering ours on both is what keeps
+    // the two sinks identical rather than merely similar.
+    sink.func("str", |v: TypedValue<TemplateTypes>| match v {
```
bugadani: We're not using this anywhere, right? `str()` is a template builtin, not something we need for expressions.

esp-template-sdk/src/process.rs
```diff
@@ -375,35 +402,56 @@ fn build_env(shared: &Rc<Shared>) -> Env {
...
+            "chip",
+            TypedValue::Struct(SomniStruct::new("Chip".into(), fields)),
+        );
     }
 
-    env.value("has_reserved_pins", shared.facts.has_reserved_pins);
+    sink.fact(
+        "has_reserved_pins",
+        TypedValue::Bool(shared.facts.has_reserved_pins),
+    );
```
bugadani: This shouldn't be part of the SDK, the template needs to be able to somehow define a list for itself. I know this conflicts with how some things are done right now, but we should maybe clean this up in a follow-up.

esp-template-sdk/src/process.rs
```diff
@@ -169,46 +125,88 @@ pub struct Vocabulary {
...
+    pub chip: Option<ChipFacts>,
     /// Known-name vocabularies backing the unknown-name hard error.
     pub vocabulary: Vocabulary,
-    /// Substitution values: spliced literally by `include_as`, and exposed to
-    /// expressions as somni values when the name is an identifier.
+    /// Substitution values: spliced into an output path, and exposed to
+    /// expressions when the name is an identifier.
     pub values: HashMap<String, FactValue>,
     /// Whether the selected module reserves any GPIOs. Backs `has_reserved_pins`.
     pub has_reserved_pins: bool,
```
bugadani: This assumes a template even has modules, which shouldn't be an SDK-hardcoded concept I think.

src/main.rs
Hahihula: the loop now evaluates every when, renders every file, and validates every output path into a `planned` vec, and only then does `fs::create_dir(&project_dir)` run and the writes the files if every path is know to be good

esp-template-sdk/src/process.rs
Hahihula: You are right; after the use of the somni-expr, this is dead code now. removing.

esp-template-sdk/src/process.rs
Hahihula: the `CHIP_FACTS` is now a `static CHIP_FACTS: LazyLock<HashMap<Chip, process::ChipFacts>>`, it builds the symbol union and for the `Chip::facts()` it uses `Arc` so we do not copy it back and forth.

esp-template-sdk/src/process.rs
Hahihula: `has_reserved_pins` is no longer a Facts field or an SDK registration. The binary sets it through `Facts::values` like `reserved_gpio_code` and `project_name` ... that should solve this

esp-template-sdk/src/process.rs
Hahihula: the `has_reserved_pins` is now set from the binary through `Facts::values` but for the broader request for the template to be able to set it's own lists I am not sure what the correct approach is. The specs suggest we have the `[vars]` but this is key:value not list ...

esp-template-sdk/src/contract.rs
```diff
@@ -118,6 +117,7 @@ const REGISTRY: &[(Version, &[(&str, FeatureKind)])] = &[(
         ("generate_parameters", FeatureKind::Value),
         ("rust_toolchain", FeatureKind::Value),
         ("reserved_gpio_code", FeatureKind::Value),
+        ("has_reserved_pins", FeatureKind::Value),
```
bugadani: This entire concept doesn't make sense to me. I can have a template that refers to these values, but a binary that doesn't provide them, and vice versa. The question is, if a binary is allowed to not provide any of these, what is the point of the facts registry?

Templates will need to be able to define their own "facts" as well (we already have `sets` for this right now). In the future, we may want a devkit/module to define a set of esp-hosted pins, for example. This needs to stay entirely in-template, or else the SDK will have to change for every new thing we think about.

And once we have the ability for a template to self-define variables, the binary could also define variables bypassing these facts. The logical conclusion to me is that we should not list optional data here at all.

Hahihula: you are right the `FeatureKind::Value` should not be touched by the sdk, either they fail in somni or during the check()

esp-template-sdk/src/process.rs
```diff
@@ -169,46 +125,84 @@ pub struct Vocabulary {
...
+    fields: Arc<IndexMap<Box<str>, FactValue>>,
+}
+
+impl ChipFacts {
+    /// Build the field map from the SDK-named scalars and the chip's symbols.
+    pub fn new(
+        name: impl Into<String>,
+        rust_target: impl Into<String>,
+        dram2_uninit_size: u64,
+        wokwi_board: impl Into<String>,
```
bugadani: Can we remove wokwi_board from the contract? It's way too specific, and the builtin template can include a bunch of `if`s to render the correct value.

I'm also on the fence about `dram2_uninit_size` - it comes from metadata, yes, but it's a rather specific piece of data from the memory map. For the moment, I don't have a good idea what to do with it (except, why is it `u64`?)

esp-template-sdk/src/template.rs
bugadani, line 103: This is unused.

esp-template-sdk/src/process.rs
```diff
@@ -373,37 +393,29 @@ fn build_env(shared: &Rc<Shared>) -> Env {
...
+/// Not "for each value, global-replace": `values` is a `HashMap`, so that walks
+/// keys in random order and a value containing `{...}` would expand or not
+/// depending on which came first. One pass also means a substituted value is
+/// never rescanned, so expansion can't chain.
 ///
-/// An unknown name is left verbatim, matching the "override or keep the
-/// default" idiom; an unbalanced `{` is copied through.
+/// An unknown name is left verbatim (the override-or-default idiom); an
+/// unbalanced `{` is copied through.
 fn interpolate(template: &str, values: &HashMap<String, FactValue>) -> String {
```
bugadani: This is essentially a second template language, why do we have it?

esp-template-sdk/src/process.rs
Hahihula: the u64 is from somni Integer which is u64,
i removed the wokwi board as that is duplicate here but the dram2_uninit_size comes from metadata curently it should be handled as the other metadata facts

esp-template-sdk/src/template.rs
Hahihula: The actual check for the probe-rs will land in the check PR, but you are right that we should include it in the template in this one.

esp-template-sdk/src/process.rs
Hahihula: This was used for filepaths and should be somni too. Fixed.

esp-template-sdk/src/process.rs
bugadani: I mean, we may want to generalize it as `memory_map.something` in case we'll add more. But we can now also evaluate an `end_address - start_address` style expression, too, so we have options to not bake in a single address in a special way.

esp-template-plugin-chip/src/lib.rs
```diff
@@ -0,0 +1,421 @@
+//! Espressif chip data, as a versioned [`TemplatePlugin`].
+//!
+//! Everything a template can reference about the target chip lives here: the
+//! chip list, the `chip` selection-group fragment, and the `chip.…` facts drawn
+//! from `esp-metadata-generated`.
+//!
+//! **This crate's version is the vocabulary.** It tracks the
+//! `esp-metadata-generated` it wraps, so a template pinning `chip = "0.4.0"` is
+//! pinning a symbol set, not a release date. That is only true because the two
```
bugadani: > not a release date.

lol, who even considers that?

esp-template-sdk/src/plugin.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn name(&self) -> &str;
+
+    /// The version a template pins. Should track whatever data this plugin
+    /// wraps, so the version identifies a vocabulary rather than a release.
+    fn version(&self) -> Version;
+
+    /// The option-tree fragment for `!Include plugin:<name>`, if any.
+    fn fragment(&self) -> Option<String> {
+        None
+    }
```
bugadani: 🤔 We could probably allow multiple fragments in a plugin (so this could take a fragment name, and we could have `Include! plugin:chip:chip_list` or something), because why not?

esp-template-sdk/src/template.rs
```diff
@@ -330,10 +329,12 @@ where
...
 
-            // The SDK's own fragments come first, so a template file cannot
-            // shadow one by sitting at the same path.
-            let contents = match builtins.resolve(&path)? {
-                Some(generated) => generated,
+            // A plugin fragment is resolved before the loader, so a template
+            // file cannot shadow one by sitting at the same path.
+            let contents = match path.strip_prefix(PLUGIN_SCHEME) {
+                Some(name) => plugins.fragment(name).ok_or_else(|| {
+                    format!("`{path}` names no plugin this template declared in `[plugins]`")
```
bugadani: That's not right - the plugin may exist, but the requested fragment can be None.

src/lib.rs
```diff
@@ -1,152 +1,30 @@
...
+pub mod manifest;
+pub mod source;
+
+pub use esp_template_sdk::{config, contract, plugin, process, template};
+pub use source::TemplateSource;
 
-pub use esp_template_sdk::{config, contract, process, template};
+/// The chip vocabulary this binary offers, re-exported so the rest of the
+/// generator doesn't care which plugin version it came from.
+pub use esp_template_plugin_chip::{Chip, ChipPlugin};
```
bugadani: `Chip` shouldn't be reexported, and I don't think `ChipPlugin` needs to be, either - I didn't find anything using it except tests, but once we have multiple versions of the chip plugin, its data shouldn't leak to the binary.

Hahihula: You are completely right. That was a leak.

src/main.rs
```diff
@@ -26,37 +27,129 @@ use std::{
...
-    .expect("failed to load bundled template");
+    let resolved =
+        resolved().expect("the bundled template must declare plugins this binary provides");
+
+    let template =
+        Template::load(root_yaml, resolved, load_include).expect("failed to load bundled template");
 
-    // The `chip` category is authored in YAML but must stay aligned with the
-    // `Chip` enum for the generator to work.
+    // A hand-authored `chip` category still has to name chips this binary knows.
```
bugadani: I wonder if this check is necessary - it still leaks chip info into the binary, when the entire chip selection is internal to the plugin. "Hand-authored" chip sections conflict with the plugin, don't they? This check will also need to know about the exact version of the chip plugin, and whether it is loaded at all, so this won't work for external templates.

esp-template-plugin-chip/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+                    {list}"#
+        )
+        .unwrap();
+    }
+
+    let mut has_reserved = false;
+    if restricted.clone().next().is_some() {
+        has_reserved = true;
+        let plucker = restricted
+            .map(|pin| format!("    let _ = peripherals.GPIO{};", pin.number))
```
bugadani:
```suggestion
            .map(|pin| format!("    let _gpio{0} = peripherals.GPIO{0};", pin.number))
```

https://github.com/esp-rs/esp-generate/pull/367

esp-template-sdk/src/process.rs
```diff
@@ -169,46 +123,69 @@ pub struct Vocabulary {
...
-    /// `chip`, `rust_target`, etc.
+    /// One field of a plugin namespace, e.g. `chip`.`rust_target`. `None` when
+    /// no resolved plugin supplied that namespace or that field.
+    pub fn field(&self, namespace: &str, field: &str) -> Option<&FactValue> {
+        self.structs.get(namespace)?.fields().get(field)
+    }
+
+    /// Insert a value, first writer wins. Binary facts go in before
+    /// template-scoped `sets`, so a template can't shadow them.
     pub fn set_value(&mut self, key: impl Into<String>, value: impl Into<FactValue>) {
```
bugadani, line 184: This is dangerous. If multiple plugins try to set the same fact, one will win silently. We should probably reject duplicates, and prefix plugin facts with `<plugin_name>.`

src/main.rs
```diff
@@ -667,11 +759,9 @@ fn main() -> Result<()> {
...
-    // Kept as a separate list rather than appended to `selected`: option and
-    // group names live in disjoint namespaces, so `option("<group>")` must not
-    // answer true (the bundled template has `coding-agent-guidance` as both a
-    // category and a group).
+    // Groups with a pick, backing `group_selected(...)`. Separate from
+    // `selected` because option and group names are disjoint namespaces —
+    // `coding-agent-guidance` is both a category and a group.
     let mut selected_groups: Vec<String> = Vec::new();
     for name in &selected {
         let (_, option) = find_option(name, &flat_options).unwrap();
```
bugadani: Arguably, we should bubble up this error instead of unwrapping it.

esp-template-sdk/src/config.rs
bugadani, line 433: Even though this is preexisting, this is the same logic as `requirements_met_against`, it would be worth deduplicating.

bugadani: This is basically `drop_unsatisfied`

esp-template-sdk/src/contract.rs
```diff
@@ -3,47 +3,36 @@
...
-/// `major.minor.patch` type that discards the suffix would compare them equal
-/// and let an rc — which may not yet implement the feature being gated —
-/// silently pass.
-///
-/// Host *tool* versions (rustc, espflash, probe-rs) are a different problem:
-/// they are reported in loose formats and their prereleases should be treated
-/// as good enough. That leniency lives in the binary's `check` module and is
-/// deliberately not shared with this type.
+/// Host *tool* versions are looser and their prereleases count as good enough;
+/// that leniency lives in the binary's `check` module, deliberately apart.
```
bugadani: It is weird that the SDK refers to the binary. The point is to separate them. We currently have one binary so this is obvious, but later it will become murky which "binary" this refers to.

esp-template-sdk/src/contract.rs
bugadani: This is only used by tests

bugadani: The check module doesn't deal with features, it checks for tool versions. This function is also only used by `min_sdk_version` which is only used by tests, so it seems redundant.

bugadani: These are also transitively used only by test code. Let's keep functions that we don't need out of the public API, unless we actually have a use for them.

esp-template-sdk/src/process.rs
```diff
@@ -324,32 +301,56 @@ impl Shared {
...
-        };
+        sink.fact(name.as_str(), typed(value));
     }
 
     let ctx = shared.clone();
-    env.function("option", move |name: &str| -> bool {
-        let vocab = &ctx.facts.vocabulary.options;
+    sink.func("option", move |name: &str| {
+        let vocab = &ctx.vocabulary.options;
         if !vocab.is_empty() && !vocab.contains(name) {
```
bugadani: Why do we allow any name if the vocabulary is empty?

bugadani (commented):
Getting there, but there are a few places that are unnecessary. Please run `cargo fmt --all`. Most of my findings were flagged by Opus, I recommend iterating a few rounds with a review agent.

esp-template-sdk/src/process.rs
Hahihula: Thanks. it should not be allowed. Switched to Option<> so now the empty and None are different cases.

esp-template-sdk/src/contract.rs
Hahihula: ok made feature private

esp-template-sdk/src/process.rs
Hahihula: Rejecting duplicates done by returning Result.

Hahihula:
> Getting there, but there are a few places that are unnecessary. Please run `cargo fmt --all`. Most of my findings were flagged by Opus, I recommend iterating a few rounds with a review agent.

Only thing flagged by 'cargo fmt --all' is in the tui.rs code i do not touched so i didn't want to shadow git history, but with this amount of changes it's probably not important, I am adding the formating fix.

bugadani (approved):
Let's call this ready, I think most of the important bits have been fixed, and we have v2 if we left in something :)

src/main.rs
```diff
@@ -821,42 +845,52 @@ fn main() -> Result<()> {
         );
     }
 
+    // Before rendering, so an existing directory fails immediately.
     fs::create_dir(&project_dir)?;
```
MabezDev: create_dir still runs before the plan loop, so a failed `when` or `as` leaves an empty project directory that then blocks a retry with "already exists". Check the path is free first and create it only after `planned` is filled.

PR #363, opened 2026-08-24
fix(template): bound-check linker error-script args before indexing

### LINKED ISSUE #358: Argument bounding in template build is off-by-one
https://github.com/esp-rs/esp-generate/blob/8718812667d41c279ca41bd957844f5b169bc06b/template/build.rs#L81

In this code from `build.rs`
```
 if args.len() > 1 {
        let kind = &args[1];
        let what = &args[2];
```

Accessing `args[2]` requires `args.len() >2`

MsfPablo:
`linker_be_nice` (in `template/build.rs`) guards its helpful-error branch with `args.len() > 1`, then indexes `args[2]` for the `what` argument. With exactly 2 arguments the guard passes but `args[2]` is out of bounds, so the linker error-handling script panics instead of printing a hint.

Guard on `args.len() > 2` instead, so the branch only runs when both `kind` (`args[1]`) and `what` (`args[2]`) are present. With fewer arguments the script now falls through to emitting the `cargo:rustc-link-arg` lines as before.

Closes #358.

### DIFF
--- a/template/build.rs
+++ b/template/build.rs
@@ -78,7 +78,10 @@ fn cargo_linker_env_var(target: &str) -> String {
 //ENDIF
 fn linker_be_nice() {
     let args: Vec<String> = std::env::args().collect();
-    if args.len() > 1 {
+    // The helpful-error branch reads both args[1] (kind) and args[2] (what),
+    // so it needs at least 3 args. Guarding on `> 1` left args[2] out of bounds
+    // when the linker wrapper was invoked with exactly 2 arguments.
+    if args.len() > 2 {
         let kind = &args[1];
         let what = &args[2];
 


SergioGasquez (commented):
Thanks for the fix! Do you mind adding a small changelog entry?

bugadani:
This isn't actually a fix. We expect the linker to call us with 3 arguments. If it doesn't, that's a bug (i.e. a specific error we're not handling, but shouldn't ignore) that will be masked by this change.

I would expect a lot more than an LLM-generated comment to close the referenced issue. If we branch on `len > 2`, we have to also provide a message for the `len > 1` case why we don't actually do what we are supposed to be doing.

## lld

https://lld.llvm.org/error_handling_script.html currently lists that both tags has a second argument.

## ld

https://man7.org/linux/man-pages/man1/ld.1.html

> Currently
           however only two kinds of error are supported: missing symbols
           and missing libraries.  Two arguments will be passed to
           script: the keyword "undefined-symbol" or `missing-lib" and
           the name of the undefined symbol or missing library.

PR #366, opened 2026-08-25
Update template for esp-hal 1.2.2

SergioGasquez:


template/Cargo.toml
```diff
@@ -21,7 +21,7 @@ test = false
 #ENDIF
 
 [dependencies]
-esp-hal = { version = "~1.1.0", features = [
+esp-hal = { version = "1.2.0-rc.0", git = "https://github.com/esp-rs/esp-hal", rev = "160b10794227eb84805b8676fe188c1110801e9d", features = [
```
bugadani: 😱 so we really need that esp-radio release, as well.

template/CODING_AGENT_GUIDANCE.md
```diff
@@ -34,10 +34,10 @@ This project was generated by `esp-generate` for Espressif MCUs.
...
 - esp-generate: https://github.com/esp-rs/esp-generate
 - esp-config: https://docs.espressif.com/projects/rust/esp-config/latest/esp_config/
-#REPLACE 1.1.0 esp-hal-version-full
-- esp-hal docs: https://docs.espressif.com/projects/rust/esp-hal/1.1.0/index.html
-#REPLACE 1.1.0 esp-hal-version-full && esp32c6 mcu
-- esp-hal additional configuration: https://docs.espressif.com/projects/rust/esp-hal/1.1.0/esp32c6/esp_hal/index.html#additional-configuration
+#REPLACE 1.2.0-rc.0 esp-hal-version-full
+- esp-hal docs: https://docs.espressif.com/projects/rust/esp-hal/1.2.0-rc.0/index.html
+#REPLACE 1.2.0-rc.0 esp-hal-version-full && esp32c6 mcu
+- esp-hal additional configuration: https://docs.espressif.com/projects/rust/esp-hal/1.2.0-rc.0/esp32c6/esp_hal/index.html#additional-configuration
```
bugadani, line 40: Is this change meaningful? :)

SergioGasquez: Not really, but it was to make with search and replace

bugadani: Yes but what's the point of the REPLACE directive if we just do it ourselves?

template/Cargo.toml
MabezDev: Why? Because we need to update esp-radio from the 1.1rc release? Can we do that in a patch release? 🤔

template/Cargo.toml
```diff
@@ -166,7 +168,7 @@ smoltcp = { version = "0.13.0", default-features = false, features = [
     #ENDIF
 ] }
 #ENDIF wifi
-esp-radio = { version = "0.18.0", features = [
+esp-radio = { version = "1.0.0-beta.0", git = "https://github.com/esp-rs/esp-hal", rev = "160b10794227eb84805b8676fe188c1110801e9d", features = [
```
MabezDev: This should still be 0.18

SergioGasquez: I tried keeping` 0.18.0,` but Cargo complains because revision `160b107…` declares `esp-radio` as `1.0.0-beta.0`

MabezDev: You're right, my bad. So let's hold this until full 1.2 is out, on the same day I'll cut esp-radio rc0 which we can use here.

SergioGasquez:
Reminder to ourselves: We need to add "Check esp32s31" job in required

Edit: Updated

template/Cargo.toml
```diff
@@ -21,7 +21,7 @@ test = false
 #ENDIF
 
 [dependencies]
-esp-hal = { version = "~1.1.0", features = [
+esp-hal = { version = "1.2.1", features = [
```
MabezDev:
```suggestion
esp-hal = { version = "~1.2.1", features = [
```

MabezDev (changes requested):

MabezDev (approved):

PR #368, opened 2026-09-02
Feat/enable external template

Hahihula:
Tested against https://github.com/Hahihula/esp-generate-templates/pull/1
Produced a result which compiles and flashes.


src/main.rs
```diff
@@ -406,15 +392,109 @@ fn check_for_update(name: &str, version: &str) {
...
+        "⚠️  Generating from the external template at `{}`. A template controls \
+         what code and dependencies end up in your project — only use ones you trust.",
+        root.display()
+    );
+
+    Ok((TemplateSource::Directory(root), checkout))
+}
+
+/// The `-o` help text. Built while clap builds `Args`, i.e. before any argument
+/// is read — so, like [`about`], it can only describe the bundled template.
```
bugadani: This will be a problem. This is meant as an accessibility function, because screen readers have trouble reading the TUI. We'll need to be able to somehow list out a template's options.

Hahihula: Now, if the '--template' is used the hellptext is generated according the template.

src/fetch.rs
```diff
@@ -0,0 +1,324 @@
...
+    let url = if base.contains("://") || base.starts_with("git@") {
+        base.to_string()
+    } else if base.split('/').count() == 2 && !base.starts_with('/') {
+        format!("https://github.com/{base}")
+    } else {
+        bail!(
+            "`--template {value}` is not an existing directory, and does not look like a \
+             repository (expected `owner/repo[@ref]`, an `https://` URL, or `git@host:path`)"
+        );
+    };
```
bugadani, line 66: So, I can point to a folder in a repository with a full URL, but only to a repository with the shorthand? We've had ideas to actually provide a preview of the builtin template in esp-rs/esp-hal, and that would be a useful feature, but it looks like the intent here is to have the entire repository be the template. Can we iterate on this a bit?

cc @MabezDev

Hahihula: Currently the logic is it tries to 'find_template_root' in the repository. It is searching for folder with the manifest file. This is meant for the case of the esp-hal where the template will live in some subfolder. But it does not cover case when there are several templates in the repository. Currently, it bails out in that case, but we can easily add the possibility for the user to choose of the multiple templates in that case from TUI if that would be prefered solution. That would enable us to have repositories with multiple templates.

MabezDev: It's quite likely we'll end up with more than one template in a repo, but I didn't spec this out in the initial plan I was working on. As Petr said, there isn't anything from adding this later, I suppose the only thing we need to maybe thing about now is the syntax, as we may box ourselves into a corner if we don't think about it at least a little bit :D.

So I'm okay with punting this on a bit.

Hahihula: The manifest specifies the template name so the low hanging fruit is just to crawl the repository list all manifests and offer the user to choose the template by name in case of more than one template name. I just felt I am already stretching the specification to its limits.

MabezDev: Agreed, lets punt this to the future.

PR #369, opened 2026-09-09
check command

Hahihula:


.github/workflows/ci.yml
```diff
@@ -111,7 +111,10 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       - name: Run tests
-        run: cargo test
+        run: cargo test --workspace
+
+      - name: Check the bundled template renders for every option
```
MabezDev, line 116: Is this the check that takes an hour? If so, we shouldn't run it unconditionally on each PR. 

We have a nightly workflow that does something similar to this btw (unless I'm misunderstanding) - have we duplicated some work here?

MabezDev: https://github.com/esp-rs/esp-generate/blob/ff9192e743bc15dff6baf20658ec42e9180de859/.github/workflows/ci.yml#L81-L82

Hahihula: the `cargo test --workspace`  runs the forgoten tests from the new crates 
the `cargo run -- check` generates the 242 template combinations but does not build them, that does the xtask. 
the `cargo run -- check` takes ~40s locally

esp-template-sdk/src/sweep.rs
```diff
@@ -0,0 +1,591 @@
...
+                out,
+            ),
+        }
+    }
+}
+
+/// Select `option` and, recursively, everything its `requires` names. One the
+/// selection cannot support is skipped rather than forced.
+fn select_with_dependencies(config: &mut ActiveConfiguration, option: &str) -> Result<(), String> {
+    let (idx, found) = find_option(option, &config.flat_options)
```
MabezDev: `find_option` returns the first of the two `probe-rs` rows in https://github.com/esp-rs/esp-generate/blob/bcd7e555cc39bf578ee1842b0c784b437c9a1d11/template/template.yaml#L73-L102, so `check --dry-run` never selects `probe-rs` on `esp32`, `esp32c2`, `esp32c3` or `esp32s2`.

Pick the variant that matches the current base, as `find_option` already documents after prune: https://github.com/esp-rs/esp-generate/blob/bcd7e555cc39bf578ee1842b0c784b437c9a1d11/esp-template-sdk/src/config.rs#L630-L642

esp-template-sdk/src/sweep.rs
```diff
@@ -0,0 +1,591 @@
...
+    flat: &[GeneratorOption],
+    pinned: &[String],
+) -> Result<Vec<Vec<String>>, String> {
+    for name in pinned {
+        if find_option(name, flat).is_none() {
+            return Err(format!("template has no option `{name}`"));
+        }
+    }
+
+    let mut picks: Vec<Vec<String>> = vec![
```
MabezDev: Pinned names are copied into every combination as written, so they skip `select_with_dependencies` and `check -o wifi` keeps `wifi` without `alloc` and still places `wifi` on `esp32h2`.

Run pinned names through the same selection path as the pool, and drop a base that `is_option_active` rejects.

src/validate.rs
```diff
@@ -0,0 +1,294 @@
...
+        std::fs::create_dir_all(out_path.parent().unwrap())?;
+        std::fs::write(out_path, contents)?;
+    }
+
+    cargo(&project, &["check"])?;
+    if selected.iter().any(|o| o == "embedded-test") {
+        cargo(&project, &["test", "--no-run"])?;
+    }
+    cargo(&project, &["clippy", "--no-deps", "--", "-Dwarnings"])?;
+    cargo(&project, &["fmt", "--", "--check"])?;
```
MabezDev, line 178: `--build` writes the rendered files and then runs `cargo fmt -- --check` without the formatting that generation applies first (`group_imports=StdExternalCrate`, `imports_granularity=Module`, and taplo on `Cargo.toml`): https://github.com/esp-rs/esp-generate/blob/bcd7e555cc39bf578ee1842b0c784b437c9a1d11/src/main.rs#L953-L975

Apply that same formatting before the check, or this gate fails on output that `esp-generate --headless` would accept.

MabezDev (changes requested):

esp-template-sdk/src/sweep.rs
Hahihula: lookups now match on name and is_option_compatible, thanks for the catch

esp-template-sdk/src/sweep.rs
Hahihula: the base now goes through the same selection path

src/validate.rs
Hahihula: extracted that formatting into a shared render::format_project

MabezDev (approved):
Thanks!

